PR #8, opened 2025-09-04
Fix Function is null error when renaming function parameters

JeffmeisterJ:
This PR fixes an error in the Ghidra Sententia plugin which occurs when renaming function variables. This happens because we don't filter the SYMBOL_RENAMED event properly.

### DIFF
--- a/sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
+++ b/sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
@@ -172,7 +172,7 @@ public class SententiaPlugin extends ProgramPlugin implements DomainObjectListen
 		while (it.hasNext()) {
 			DomainObjectChangeRecord evt = it.next();
 			// Check if we're dealing with a rename and the is applied to a function
-			if (evt.getEventType().getId() == ProgramEvent.SYMBOL_RENAMED.getId() && ((ProgramChangeRecord)evt).getObject().getClass() == FunctionSymbol.class) {				
+			if (evt.getEventType().getId() == ProgramEvent.SYMBOL_RENAMED.getId() && ((ProgramChangeRecord)evt).getObject() instanceof FunctionSymbol) {				
 				
 				Function changedFunction = currentProgram.getFunctionManager().getFunctionAt(((ProgramChangeRecord)evt).getStart());
 				
--- a/sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
+++ b/sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
@@ -177,8 +177,7 @@ public class SententiaPlugin extends ProgramPlugin implements DomainObjectListen
 				Function changedFunction = currentProgram.getFunctionManager().getFunctionAt(((ProgramChangeRecord)evt).getStart());
 				
 				try {
-					FunctionDescriptor changedFunctionDescriptor = new FunctionDescriptor(changedFunction);
-					sententiaAPI.addSignatureToDB(changedFunctionDescriptor);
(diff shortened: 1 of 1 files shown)


sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
```diff
@@ -170,10 +171,13 @@ public void domainObjectChanged(DomainObjectChangedEvent event) {
 		Iterator<DomainObjectChangeRecord> it = event.iterator();
 		while (it.hasNext()) {
 			DomainObjectChangeRecord evt = it.next();
-			if (evt.getEventType().getId() == ProgramEvent.SYMBOL_RENAMED.getId()) {				
+			// Check if we're dealing with a rename and the is applied to a function
```
akaIDIOT, line 174: *and the is applied*? Should that be "that is applied"?

sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
```diff
@@ -170,10 +171,13 @@ public void domainObjectChanged(DomainObjectChangedEvent event) {
...
+			if (evt.getEventType().getId() == ProgramEvent.SYMBOL_RENAMED.getId() && ((ProgramChangeRecord)evt).getObject() instanceof FunctionSymbol) {				
+				
+				Function changedFunction = currentProgram.getFunctionManager().getFunctionAt(((ProgramChangeRecord)evt).getStart());
 				
 				try {
-					FunctionDescriptor changedFunctionDescriptor = new FunctionDescriptor(currentProgram.getFunctionManager().getFunctionAt(((ProgramChangeRecord)evt).getStart()));
+					FunctionDescriptor changedFunctionDescriptor = new FunctionDescriptor(changedFunction);
 					sententiaAPI.addSignatureToDB(changedFunctionDescriptor);
```
akaIDIOT, line 181: We could ask ourselves whether the additional local makes things more readable, would this make sense in a single go?


```suggestion
					sententiaAPI.addSignatureToDB(new FunctionDescriptor(changedFunction));
```

sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
```diff
@@ -170,10 +171,13 @@ public void domainObjectChanged(DomainObjectChangedEvent event) {
 		Iterator<DomainObjectChangeRecord> it = event.iterator();
 		while (it.hasNext()) {
 			DomainObjectChangeRecord evt = it.next();
-			if (evt.getEventType().getId() == ProgramEvent.SYMBOL_RENAMED.getId()) {				
+			// Check if we're dealing with a rename and the is applied to a function
+			if (evt.getEventType().getId() == ProgramEvent.SYMBOL_RENAMED.getId() && ((ProgramChangeRecord)evt).getObject() instanceof FunctionSymbol) {
```
akaIDIOT, line 175: Not familiar with Ghidra's API, I'm assuming this is the recommended way of checking / filtering this? :)

akaIDIOT (approved):

sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
JeffmeisterJ: I think so, but the API docs are pretty meager. Seems to be the most specific solution to this problem though.

sententia/src/main/java/nl/nfi/sententia/SententiaPlugin.java
JeffmeisterJ: Yes, it was a bit late and apparently my brain had turned to goo 🙃

PR #11, opened 2025-10-09
Restrict sentence-transformers to 4.*, re-lock dependencies

akaIDIOT:
See #10, should provide part 1 of the proposed 2-step fix 🙃


JeffmeisterJ (approved):
Seems good as a temporary fix, while we wait for updating to v5.

PR #12, opened 2025-10-09
ATF example

nilswiersma:
PR to add an examples folder for now only populated with a database and ghidra project previously created by @JeffmeisterJ


JeffmeisterJ (approved):
Looks great, thanks for writing up the example!

PR #14, opened 2026-04-20
Enable gha for sub-project citatio

akaIDIOT:


citatio/pyproject.toml
```diff
@@ -15,7 +15,7 @@ authors = [
     {name = "Judith van de Wetering", email = "22766364+judithvdw@users.noreply.github.com"},
     {name = "Anne Fleur van Luenen", email = "50952036+fleurvanl@users.noreply.github.com"},
 ]
-requires-python = ">=3.12"
+requires-python = ">=3.13,<3.14"
```
ranieri, line 18: Unfortunate that we can't use 3.14 yet. Let's keep an eye on it.

ranieri (approved):

citatio/pyproject.toml
akaIDIOT: Yeah lets; I'd like to have things `>=3.13`, but the dependency resolution was both slow and iffy... 

I'll try to run this at home on a faster machine not encumbered by PyPI filters ;)

PR #15, opened 2026-04-20
Add architecture documentation for `asmtransformers`

ranieri:
Add architecture documentation for `asmtransformers`, detailing substems and workflows

### DIFF
--- a/asmtransformers/docs/architecture.md
+++ b/asmtransformers/docs/architecture.md
@@ -11,9 +11,9 @@ This package is currently optimized for ARM64 assembly, but not every subsystem 
 
 ## Main Subsystems
 
-### 1. ISA preprocessing
+### ISA preprocessing
 
-The ISA-specific preprocessing path is implemented in [arm64.py](../asmtransformers/arm64.py).
+The ISA-specific preprocessing path is implemented in [asmtransformers.arm64](../asmtransformers/arm64.py).
 
 Its responsibilities are:
 
@@ -25,13 +25,13 @@ Its responsibilities are:
 The central type is `Preprocessor`. It accepts either:
 
 - a `dict[int, list[str]]` mapping basic-block offsets to instruction strings
-- a `networkx.DiGraph` with jTrans-style node data containing `asm`
+- a `networkx.DiGraph` with an `asm` attribute on the (basic block) graph nodes.
 
 The output is a flat token list suitable for a tokenizer or vocabulary builder.
 
-### 2. Operand normalization
+### Operand normalization
 
-Operand normalization helpers live in [operands.py](../asmtransformers/operands.py).
+Operand normalization helpers live in [asmtransformers.operands](../asmtransformers/operands.py).
 
 These helpers reduce token explosion caused by raw numeric values. The current ARM64 path uses:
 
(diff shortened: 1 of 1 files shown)


asmtransformers/docs/architecture.md
```diff
@@ -0,0 +1,129 @@
...
+`asmtransformers` is the training and inference core of the monorepo. It turns control-flow-graph representations of assembly functions into token sequences, feeds those sequences into transformer models, and exposes a script-oriented workflow for preprocessing, pretraining, finetuning, evaluation, and embedding generation.
+
+This package is currently optimized for ARM64 assembly, but not every subsystem is ARM64-specific. The main architectural distinction is:
+
+- ISA-specific logic lives in preprocessing and tokenization.
+- Model composition, training flow, and most dataset handling are reusable patterns.
+
+## Main Subsystems
+
+### 1. ISA preprocessing
```
akaIDIOT: Authors will now be responsible for keeping this numbering correct, is it worth that?

asmtransformers/docs/architecture.md
```diff
@@ -0,0 +1,129 @@
...
+
+- parsing assembly instructions into opcode and operand tokens
+- normalizing bracketed memory expressions and register sets into stable token streams
+- identifying branch instructions and rewriting concrete jump targets into relative `JUMP_ADDR_*` tokens
+- flattening a CFG-like input into a single model-ready token sequence
+
+The central type is `Preprocessor`. It accepts either:
+
+- a `dict[int, list[str]]` mapping basic-block offsets to instruction strings
+- a `networkx.DiGraph` with jTrans-style node data containing `asm`
```
akaIDIOT: The first one I can parse, the second isn't really enough information for me to understand what this means; what *does* it mean?

asmtransformers/docs/architecture.md
```diff
@@ -0,0 +1,129 @@
...
+
+Dataset helpers live in [datasets/sentencelabel.py](../asmtransformers/datasets/sentencelabel.py).
+
+`LazySentenceLabelDataset` bridges Hugging Face datasets and sentence-transformers training by:
+
+- grouping rows by label
+- lazily sampling multiple examples per label
+- emitting `InputExample` objects suitable for triplet-style training
+
+This layer is largely architecture-agnostic as long as the dataset schema remains consistent.
```
akaIDIOT, line 68: This sentence reads like a review of the repository rather than documentation. Is it meant as a warning or something else?

asmtransformers/docs/architecture.md
akaIDIOT, line 1: When talking about scripts, *naming* the reference after a file name makes sense. When it's talking about a module, would *naming* them after the module not make more sense? 

E.g.: `live in [operands.py](../asmtransformers/operands.py).` → `live in [asmtransformers.operands](...)`

akaIDIOT (approved):

asmtransformers/docs/architecture.md
ranieri: Gooi point!

asmtransformers/docs/architecture.md
ranieri: One of the aims of this document is to explain to developers what bits are ISA-specific and which bits are not. The point of sentence is to tell developers this bit is not ISA-specific, and just relies on the data shape remains the same.

asmtransformers/docs/architecture.md
ranieri: jTrans stores the CFGs in networkx.DiGraph objects. In these objects, every node is a basic block  with an additional `asm` attribute storing the code of the basic block. Our process function can take these objects as input and then immediately converts them to out preferred `dict[int, list[str]]`. 

I don't know if this code path is still used for anything. I can drop it from the docs, but we should really figure out whether we want to keep it in the code as well.

PR #16, opened 2026-04-21
Bugfix: we were passing model_args in the wrong (positional) parameter

ranieri:
Fixed and made some args kwarg-only to avoid repeat of the issue.


akaIDIOT (approved):
👏🏻

PR #18, opened 2026-04-21
Fix `__len__` in LazySentenceLabelDataset and add corresponding test

ranieri:
We only yield N samples per category. The current __len__() returns the total number of eligible samples.


akaIDIOT (approved):
I believe you 😅

PR #20, opened 2026-04-21
[codex] Fix sentence-transformers 5 compatibility

ranieri:
## Summary
- update the custom asm sentence-transformers wrapper to follow the sentence-transformers 5.x input module contract
- replace the broken pretokenized tokenizer path with ids-first encoding plus tokenizer-managed padding
- restore tied embedding weights from checkpoint load so ARM64BERT embeddings remain stable across reloads

## Root Cause
sentence-transformers 5.x changed the Transformer/tokenizer contract, and the upstream transformers loader now leaves the tied word embedding parameter uninitialized for this checkpoint layout.

## Validation
- `pdm run test`

### DIFF
--- a/asmtransformers/asmtransformers/models/asmbert.py
+++ b/asmtransformers/asmtransformers/models/asmbert.py
@@ -21,7 +21,10 @@ class ASMBertModel(BertModel):
     BinBert shares parameters between the position embeddings and the jump target embeddings.
     """
 
-    _tied_weights_keys = [*(BertModel._tied_weights_keys or []), 'embeddings.position_embeddings.weight']
+    _tied_weights_keys = {
+        **(BertModel._tied_weights_keys or {}),
+        'embeddings.position_embeddings.weight': 'embeddings.word_embeddings.weight',
+    }
 
     def __init__(self, config, add_pooling_layer=True, *, delay_tie_for_load=False):
         super().__init__(config, add_pooling_layer)
@@ -74,7 +77,10 @@ class ASMBertForMaskedLM(BertForMaskedLM):
     Additionally, BinBert shares parameters between the position embeddings and the jump target embeddings.
     """
 
-    _tied_weights_keys = [*BertForMaskedLM._tied_weights_keys, 'bert.embeddings.position_embeddings.weight']
+    _tied_weights_keys = {
+        **BertForMaskedLM._tied_weights_keys,
+        'bert.embeddings.position_embeddings.weight': 'bert.embeddings.word_embeddings.weight',
+    }
 
     def __init__(self, config):
         """Override BertForMaskedLM's init to add Jump Target Prediction
--- a/asmtransformers/asmtransformers/models/asmbert.py
+++ b/asmtransformers/asmtransformers/models/asmbert.py
@@ -3,8 +3,6 @@ import warnings
 from copy import deepcopy
 
 import torch
-from torch import nn
-from torch.nn import CrossEntropyLoss
 from transformers import BertForMaskedLM, BertModel, BertTokenizer
 from transformers.modeling_outputs import MaskedLMOutput
 from transformers.models.bert.modeling_bert import BertOnlyMLMHead, BertPreTrainedModel
@@ -34,7 +32,7 @@ class ASMBertModel(BertModel):
             # Keep a checkpoint-shaped positional table around just for loading
             # legacy checkpoints that only store the shared embedding under the
             # position-embedding key.
-            self.embeddings.position_embeddings = nn.Embedding(config.vocab_size, config.hidden_size)
+            self.embeddings.position_embeddings = torch.nn.Embedding(config.vocab_size, config.hidden_size)
         else:
             self.tie_shared_embeddings()
 
@@ -165,7 +163,7 @@ class ASMBertForMaskedLM(BertForMaskedLM):
         )
         sequence_output = outputs[0]
 
-        loss_fct = CrossEntropyLoss()  # -100 index = padding token
+        loss_fct = torch.nn.CrossEntropyLoss()  # -100 index = padding token
 
         # Masked Language Modelling
         prediction_scores = self.cls(sequence_output)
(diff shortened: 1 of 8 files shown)


citatio/pyproject.toml
akaIDIOT, line 1: We can lift this change to main, right? This need not be part of sentence-transformers 4 → 5 🤔

ranieri:
I've tested `pretrain.py` is functional. 
I know `finetune.py` is broken.

asmtransformers/asmtransformers/models/asmbert.py
```diff
@@ -1,7 +1,9 @@
 import json
+import warnings
 from copy import deepcopy
 
 import torch
+from torch import nn
 from torch.nn import CrossEntropyLoss
```
akaIDIOT: These two can likely be combined, I'd expect.

asmtransformers/asmtransformers/models/asmbert.py
```diff
@@ -20,15 +22,59 @@ class ASMBertModel(BertModel):
...
     """
 
-    _tied_weights_keys = [*(BertModel._tied_weights_keys or []), 'embeddings.position_embeddings.weight']
+    _tied_weights_keys = {
+        **(BertModel._tied_weights_keys or {}),
+        'embeddings.position_embeddings.weight': 'embeddings.word_embeddings.weight',
+    }
 
-    def __init__(self, config, add_pooling_layer=True):
+    def __init__(self, config, add_pooling_layer=True, *, delay_tie_for_load=False):
```
akaIDIOT, line 28: KW-only is great, should we include `add_pooling_layer` with this, or is it required to keep that around as a positional for something else?

asmtransformers/asmtransformers/models/asmbert.py
```diff
@@ -20,15 +22,59 @@ class ASMBertModel(BertModel):
...
+                UserWarning,
+                stacklevel=2,
+            )
+            model.embeddings.word_embeddings.weight.data.copy_(model.embeddings.position_embeddings.weight.data)
+            loading_info['missing_keys'].discard('embeddings.word_embeddings.weight')
+        model.tie_shared_embeddings()
+
+        if output_loading_info:
+            return model, loading_info
+        return model
```
akaIDIOT, line 73: For the record: I hate this construction, but it seems part of `super()`'s contract, so ... fine 😤

asmtransformers/asmtransformers/models/asmbert.py
```diff
@@ -198,26 +247,26 @@ def set_tokenizer(self, tokenizer):
...
-            truncation=True,
-            padding=True,
+            encoded_inputs.append(
+                {
+                    # The assembly preprocessor already splits the function into model vocabulary tokens.
+                    'input_ids': self.tokenizer.convert_tokens_to_ids(tokens[:512]),
+                }
+            )
+
+        return self.tokenizer.pad(
```
akaIDIOT, line 261: As this is calling `pad()`, is the `+= self.padding[...]` above needed? What's the difference?

asmtransformers/asmtransformers/models/asmsentencebert.py
```diff
@@ -26,30 +27,58 @@ def __init__(
...
         nn.Module.__init__(self)
 
-        if model_args is None:
-            model_args = {}
-
-        model = ASMBertModel.from_pretrained(model_name_or_path, **model_args)
-        model.tokenizer = tokenizer
-        self.tokenizer = tokenizer
-
+        model = ASMBertModel.from_pretrained(model_name_or_path, **(model_args or {}))
```
akaIDIOT, line 30: This is better than the `if is None` we had 👌🏻

asmtransformers/asmtransformers/models/asmsentencebert.py
```diff
@@ -26,30 +27,58 @@ def __init__(
...
+
+    @property
+    def tokenizer(self):
+        return self.processor
+
+    @tokenizer.setter
+    def tokenizer(self, tokenizer):
+        self.processor = tokenizer
+        if hasattr(self, 'model'):
+            self.model.tokenizer = tokenizer
```
akaIDIOT, line 81: Considering ST5 has renamed these things: what code still needs there to be a `tokenizer` attribute, our own or someone else's? (in the second case: could we just ditch the property?)

asmtransformers/asmtransformers/models/asmbert.py
```diff
@@ -198,26 +247,26 @@ def set_tokenizer(self, tokenizer):
...
-
-        tokenized = self.tokenizer(
-            tokens_batch,
-            # input is pre-tokenized by the preprocessor
-            is_split_into_words=True,
-            add_special_tokens=False,
-            # truncate the encoded sequence to a maximum of max_length tokens
-            truncation=True,
-            padding=True,
+            encoded_inputs.append(
```
akaIDIOT, line 254: Not a fan of the init + append style here, but the old code did that too. Let's fix that later.

akaIDIOT (approved):

asmtransformers/asmtransformers/models/asmbert.py
ranieri: That's part of the `BertModel.__init__` contract.

asmtransformers/asmtransformers/models/asmsentencebert.py
ranieri: A lot of the logic surrounding the tokenizer can be simplified substantially, but I'm leaving this for a future PR

asmtransformers/asmtransformers/models/asmbert.py
ranieri: Good catch! The padding tokes we were adding manually are also interfering with the attention mask.

ranieri: Further inspection reveals this was present in main as well. I'm making an issue and leaving it for a future PR

asmtransformers/asmtransformers/models/asmbert.py
drxzcl: Done!

PR #24, opened 2026-04-22
Add unit tests for ASMBert with tied embeddings and safe serialization

ranieri:
Note: `save_pretrained()` default changed somewhere between HF Transformers version 4.34 and 4.38 to prefer safe serialisation. This serialization method cannot deal with our way of sharing weights, so we can only currenltly save using explicit `safe_serialization=False`. 

This PR codifies existing behaviour only.

### DIFF
--- a/asmtransformers/tests/test_asmbert.py
+++ b/asmtransformers/tests/test_asmbert.py
@@ -1,0 +1,114 @@
+import torch
+import pytest
+from transformers import BertConfig
+
+from asmtransformers.models.asmbert import ASMBertForMaskedLM
+
+
+def create_model():
+    torch.manual_seed(0)
+    config = BertConfig(
+        vocab_size=32,
+        hidden_size=16,
+        num_hidden_layers=1,
+        num_attention_heads=2,
+        intermediate_size=32,
+        max_position_embeddings=8,
+        hidden_dropout_prob=0.0,
+        attention_probs_dropout_prob=0.0,
+    )
+    return ASMBertForMaskedLM(config)
+
+
+def test_position_and_word_embeddings_are_tied():
+    model = create_model()
+
+    word_embeddings = model.base_model.embeddings.word_embeddings
+    position_embeddings = model.base_model.embeddings.position_embeddings
+
+    assert position_embeddings is word_embeddings
+    assert position_embeddings.weight is word_embeddings.weight
+    assert position_embeddings.weight.data_ptr() == word_embeddings.weight.data_ptr()
+
+
+def test_forward_reports_losses_and_jtp_ignores_out_of_range_labels():
+    model = create_model()
+    model.eval()
+
+    input_ids = torch.tensor([[1, 2, 3, 4]])
+    labels = torch.tensor([[7, 9, -100, 31]])
+    labels_with_different_non_jtp_targets = torch.tensor([[7, 10, -100, 30]])
+
+    output = model(input_ids=input_ids, labels=labels)
+    output_with_different_non_jtp_targets = model(
+        input_ids=input_ids, labels=labels_with_different_non_jtp_targets
+    )
+
+    assert output.loss is not None
+    assert output.masked_lm_loss is not None
+    assert output.jtp_loss is not None
+    assert torch.isclose(output.loss, output.masked_lm_loss + output.jtp_loss)
+    assert torch.isclose(output.jtp_loss, output_with_different_non_jtp_targets.jtp_loss)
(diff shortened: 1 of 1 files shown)


asmtransformers/tests/test_asmbert.py
```diff
@@ -0,0 +1,114 @@
+import torch
+import pytest
+from transformers import BertConfig
+
+from asmtransformers.models.asmbert import ASMBertForMaskedLM
+
+
+def create_model():
```
akaIDIOT, line 8: This should be a fixture, I'd think?

asmtransformers/tests/test_asmbert.py
```diff
@@ -0,0 +1,114 @@
...
+    torch.manual_seed(0)
```
akaIDIOT, line 12: If this is really needed, I'd expect a note on what's getting randomized that needs to be stabilized here.

asmtransformers/tests/test_asmbert.py
```diff
@@ -0,0 +1,114 @@
...
+    assert position_embeddings.weight.data_ptr() == word_embeddings.weight.data_ptr()
+
+
+def test_forward_reports_losses_and_jtp_ignores_out_of_range_labels():
+    model = create_model()
+    model.eval()
+
+    input_ids = torch.tensor([[1, 2, 3, 4]])
+    labels = torch.tensor([[7, 9, -100, 31]])
+    labels_with_different_non_jtp_targets = torch.tensor([[7, 10, -100, 30]])
```
akaIDIOT, line 45: The intention here is not very clear. The actual code was a while ago, but what does labels imply here? 🤔 

The second set of labels is also not very different; does the name mean that it's just different, and still no jump targets? Or does the first contain jump targets but the second one does not?

asmtransformers/tests/test_asmbert.py
```diff
@@ -0,0 +1,114 @@
...
+    optimizer.step()
+
+    position_weight = model.base_model.embeddings.position_embeddings.weight
+
+    assert position_weight is shared_weight
+    assert position_weight.data_ptr() == shared_weight.data_ptr()
+    assert not torch.equal(shared_weight_before, shared_weight.detach())
+
+
+def test_pickle_save_and_load_preserves_tied_embeddings(tmp_path):
```
akaIDIOT, line 91: I like the simplicity of this test, create-save-load-check 👏🏻 (and an important check at that).

asmtransformers/tests/test_asmbert.py
```diff
@@ -0,0 +1,114 @@
...
+def test_safe_serialization_rejects_tied_embeddings(tmp_path):
+    model = create_model()
+
+    # The embeddings are tied at runtime, but that tie is not declared through
+    # Transformers' tied-weight metadata. Safe serialization should therefore fail.
+    # See:
+    # https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.get_expanded_tied_weights_keys
+    #
+    # If the model starts declaring this tie properly, replace this with a positive
+    # safe-serialization round-trip test.
```
akaIDIOT, line 119: Fair enough. Would it make sense to override `save_pretrained` without the parameter and call super with `safe_serialization=False`? Might be overkill for a test that is intended as a regression for the upgrade in #20 (?).

akaIDIOT (approved):

ranieri:
Lol doe es niet werken zo laat, man!

asmtransformers/tests/test_asmbert.py
ranieri: I considered it. For this particular set of tests, I like to start fresh every time.

asmtransformers/tests/test_asmbert.py
drxzcl: The pair contains the same jump targets (<8, in the model instantiated up top), but different non-jump targets. The test checks that the JTP loss is the same, but the total loss is different.

This might need a line comment :P

asmtransformers/tests/test_asmbert.py
drxzcl: In an ordinary world, yes. But unsafe save is getting the same guard in ST5, so we need to do it properly anyway.

asmtransformers/tests/test_asmbert.py
drxzcl: Added comment

asmtransformers/tests/test_asmbert.py
akaIDIOT: That's ... how a fixture works 🤔 

Unless you tell it to stick around per module / session / ...

PR #25, opened 2026-04-23
Add declare tied embeddings in ASMBert models and update safe seriali…

ranieri:
…zation tests

### DIFF
--- a/asmtransformers/asmtransformers/models/asmbert.py
+++ b/asmtransformers/asmtransformers/models/asmbert.py
@@ -20,8 +20,14 @@ class ASMBertModel(BertModel):
     BinBert shares parameters between the position embeddings and the jump target embeddings.
     """
 
+    _tied_weights_keys = [*(BertModel._tied_weights_keys or []), 'embeddings.position_embeddings.weight']
+
     def __init__(self, config, add_pooling_layer=True):
         super().__init__(config, add_pooling_layer)
+        self._tie_weights()
+
+    def _tie_weights(self):
+        """Declare and re-apply the custom embedding alias for HF save/load flows."""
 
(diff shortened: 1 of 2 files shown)


asmtransformers/tests/test_asmbert.py
```diff
@@ -23,6 +22,21 @@ def create_model():
...
+        vocab_size=32,
+        hidden_size=16,
+        num_hidden_layers=1,
+        num_attention_heads=2,
+        intermediate_size=32,
+        max_position_embeddings=8,
+        hidden_dropout_prob=0.0,
+        attention_probs_dropout_prob=0.0,
+    )
+    return ASMBertModel(config)
```
akaIDIOT: As with the other PR, this should be a fixture (that will get re-recreated at every test, unless you tell pytest to keep it around for longer ;))

asmtransformers/asmtransformers/models/asmbert.py
```diff
@@ -20,8 +20,14 @@ class ASMBertModel(BertModel):
     BinBert shares parameters between the position embeddings and the jump target embeddings.
     """
 
+    _tied_weights_keys = [*(BertModel._tied_weights_keys or []), 'embeddings.position_embeddings.weight']
```
akaIDIOT, line 23: Aaah, this declaration is understood by the super types so they'll allow *the trick™* under 'safe' conditions? Noice 👌🏻

akaIDIOT (approved):

asmtransformers/asmtransformers/models/asmbert.py
ranieri: Yes! Altogether too much magic for my tastes. Also, it changes syntax in T5.

PR #27, opened 2026-04-23
Apply formatting to scripts, fix check violations

akaIDIOT:


asmtransformers/tests/test_asmbert.py
akaIDIOT, line 1: This file is part of #26, which should really be merged before this one.

(trains are hard on GitHub 🚂 😞)

akaIDIOT: #26 has now been merged 🎉

ranieri (approved):
Much better

akaIDIOT:
> Much better

Well, at least it's all the same now. And no more `datetime.datetime.now()` without a timezone 😬

PR #28, opened 2026-04-23
Force editable install on asmtransformers from citatio

akaIDIOT:


citatio/pyproject.toml
```diff
@@ -32,6 +32,9 @@ dependencies = [
 check = [
     "ruff>=0.11.9",
 ]
+dev = [
+    "-e file:///${PROJECT_ROOT}/../asmtransformers",
```
ranieri, line 36: Verified that changes in `asmtranformers` can cause `citatio` tests to fail 😅

ranieri (approved):

citatio/pyproject.toml
akaIDIOT: Yeah, that was ... unfortunate, but this should remedy the situation for the foreseeable fjoetjoere 🙏🏻

PR #30, opened 2026-04-24
Resolve double-padding issue leading to incorrect attention mask.

### LINKED ISSUE #29: Double padding affects attention mask
We combine manual padding with tokenizer.pad(). That last call sets the attention mask to all 1s regardless of the actual content length.

ranieri:
Resolves #29

### DIFF
--- a/asmtransformers/tests/test_arm64.py
+++ b/asmtransformers/tests/test_arm64.py
@@ -1,8 +1,8 @@
+import importlib.resources
 import json
 
 import pytest
 from networkx import DiGraph
-from transformers import BertTokenizer
 
 from asmtransformers import arm64
 from asmtransformers.models.asmbert import ARM64Tokenizer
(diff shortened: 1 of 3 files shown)


ranieri:
NOTE: This change affects the "golden embedding" test. We're now correctly attending the first 450 tokens, instead of 512.

asmtransformers/tests/test_arm64.py
```diff
@@ -129,3 +133,20 @@ def format_operand(self, operand):
     assert 'JUMP_ADDR_6' in tokens
     assert '0x78' not in tokens
     assert tokens.count('OBFUSCATED') == 2
+
+
+def test_arm64_tokenizer_masks_padding_tokens():
+    tokenizer_path = 'asmtransformers/models/arm64bert'
```
akaIDIOT: This is a relative path, right? Can we make that PWD-agnostic (and potentially based off of a fixture)?

asmtransformers/tests/test_asmsentencebert.py
```diff
@@ -54,9 +54,9 @@ def test_single_embedding(anchor, model):
     """We took one sample and calculated the sum, min and max of the embedding. This should be used as reference
     material to notice if anything changed over time"""
     embedding = model.encode(anchor)
-    assert np.isclose(embedding.sum(), -0.05649805)
-    assert np.isclose(embedding.min(), -0.11195006)
-    assert np.isclose(embedding.max(), 0.115316495)
+    assert np.isclose(embedding.sum(), -0.09272218)
+    assert np.isclose(embedding.min(), -0.10641833)
+    assert np.isclose(embedding.max(), 0.116405316)
```
akaIDIOT, line 59: No idea whether it's supposed to be, but it's sort of soothing that these don't change that much 🤷🏻

akaIDIOT (approved):

PR #33, opened 2026-04-24
Refactor `preprocess.py` to simplify preprocessing logic

ranieri:


akaIDIOT (approved):
That's a chunk less code, noice.

PR #36, opened 2026-04-28
Remove delayed ASMBert load path

ranieri:


akaIDIOT (approved):
I kinda like this 😍 

Big red PR, anyday 👏🏻

PR #39, opened 2026-04-30
Prepocessor for x86_64 arch

AstroVasseL:


asmtransformers/asmtransformers/x86.py
```diff
@@ -0,0 +1,227 @@
+import re
+from collections.abc import Callable, Iterator
+
+from networkx import DiGraph
```
ranieri: We are planning on exorcising this particular dependency from the arm64 branches in favour of dicts/lists.

asmtransformers/asmtransformers/x86.py
```diff
@@ -0,0 +1,227 @@
...
+    'jecxz',
+    'jrcxz',
+    # loop instructions (branch if cx != 0)
+    'loop',
+    'loope',
+    'loopne',
+)
+
+
+# Ghidra prefixes memory operands with a size qualifier followed by the keyword 'ptr'
```
ranieri: That's not just Ghidra - it's standard intel syntax x86 assembler. That's because x86 instruction can operate on different operand widths. With registers the width is implied, but in memory locations that's not specified and needs to be made explicit.

asmtransformers/asmtransformers/x86.py
```diff
@@ -0,0 +1,227 @@
+import re
+from collections.abc import Callable, Iterator
+
+from networkx import DiGraph
+
+from asmtransformers.operands import is_offset
+
+
+# based mostly on: https://en.wikipedia.org/wiki/List_of_x86_instructions
+BRANCH_INSTRUCTIONS = (
```
ranieri, line 8: Technically for flow control there's also the REP prefix, but since that doesn't modify the PC, I don't think we need to worry about it.

asmtransformers/asmtransformers/x86.py
```diff
@@ -0,0 +1,227 @@
...
+        match expr[i]:
+            case ' ':
+                i += 1
+            case '+' | '*':
+                yield expr[i]
+                i += 1
+            case '-':
+                # unary minus: attach to the number that follows when preceded by an operator or start
+                prev = expr[:i].rstrip()
+                if not prev or prev[-1] in ('+', '-', '*'):
```
ranieri, line 143: Multiplication addressing FTW!

ranieri (approved):

PR #40, opened 2026-04-30
Draft: RISC-V preprocessor

fleurvanl:


asmtransformers/tests/test_riscv.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_parse_no_operands():
+    assert riscv.parse_instruction('ret') == ('ret', ())
+    # there are only 20 items that can have a c. extension, indicating the instruction should be saved in less memory
+    # so we keep it attached to the instruction
+    assert riscv.parse_instruction('c.nop') == ('c.nop', ())
+
+
+def test_parse_single_operand():
+    assert riscv.parse_instruction('j x032') == ('j', ('x032',))
```
ranieri: is "x032" a typo?

I would expect 0x32 (address) or x31 (register, but it only goes up to x31, there is no x32).

asmtransformers/asmtransformers/riscv.py
```diff
@@ -0,0 +1,147 @@
+import re
+from collections.abc import Iterator
+
+from networkx import DiGraph
```
ranieri: This dependency is getting removed soon. We are standardizing on dict/list.

asmtransformers/tests/test_riscv.py
```diff
@@ -0,0 +1,108 @@
...
+
+    tokens = ObfuscatingTokenizer().preprocess(graph)
+    # expect a jump token towards the second basic block, but expect the other two occurrences of 0x78 to have been
+    # obfuscated by obfuscate_offset
+    assert 'JUMP_ADDR_6' in tokens
+    assert '0x78' not in tokens
+    assert tokens.count('OBFUSCATED') == 2
+
+
+# todo: adjust to risc-v; requires a riscv subfolder under asmtransformers/models
```
ranieri: This is a tokenizer thing, not a preprocessor thing, I'm OK with dropping this from here. It will be lifted from the ARM64 impl when we start generalizing.

asmtransformers/pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "torch",
     "transformers",
     "tzlocal>=5.3.1",
+    "pandas>=3.0.2",
```
ranieri: Do we need this? Who's using it?

Also, it's not in the PDM lock change. If we really want to make sure everyone gets it, we need to re-lock.

ranieri:
We're still missing immediate value bucketing/normalization. That can go in a later pull request.

ranieri (approved):

PR #41, opened 2026-04-30
Refactor ASM embedding inference

ranieri:
Switch to native (no ST) embedding/inference. Only use a thin ST adapter to enable triplet loss training.


asmtransformers/asmtransformers/models/asmsentencebert.py
```diff
@@ -1,164 +1,83 @@
...
 
     def __init__(
         self,
         model_name_or_path: str,
-        tokenizer,
         *,
-        max_seq_length: int | None = None,
-        do_lower_case: bool = False,
-        model_args: Mapping[str, Any] | None = None,
+        model_args: dict[str, Any] | None = None,
```
akaIDIOT: This went from `Mapping[str, Any]` to `dict[str, Any]`; doubt that's a better declared type, it's only splat along right?

asmtransformers/asmtransformers/models/asmsentencebert.py
```diff
@@ -1,164 +1,83 @@
...
-    @property
-    def tokenizer(self):
-        return self.processor
+    def get_embedding_dimension(self) -> int:
+        return self.model.config.hidden_size
+
+    def tokenize(self, texts, **kwargs):
+        return self.preprocess(texts, **kwargs)
+
+    def forward(self, features: dict[str, Any], **kwargs) -> dict[str, Any]:
```
akaIDIOT, line 34: Lots of `dict[...]` here too.

asmtransformers/asmtransformers/models/asmsentencebert.py
```diff
@@ -1,164 +1,83 @@
...
-        self.unpad_inputs = False
 
-    @property
-    def tokenizer(self):
-        return self.processor
+    def get_embedding_dimension(self) -> int:
+        return self.model.config.hidden_size
+
+    def tokenize(self, texts, **kwargs):
+        return self.preprocess(texts, **kwargs)
```
akaIDIOT, line 32: Simple enough, but the terminology is very confusing right now; `tokenize()` calls `preprocess()`, which in turn calls `tokenizer()`. Not sure how much of this is our doing, but there's a chance for improvement here 😬

asmtransformers/asmtransformers/models/asmsentencebert.py
```diff
@@ -1,164 +1,83 @@
...
+            model_inputs['token_type_ids'] = features['token_type_ids']
 
-    @tokenizer.setter
-    def tokenizer(self, tokenizer):
-        self.processor = tokenizer
-        if hasattr(self, 'model'):
-            self.model.tokenizer = tokenizer
+        outputs = self.model(**model_inputs)
+        features['token_embeddings'] = outputs.last_hidden_state
+        return features
```
akaIDIOT, line 47: I'm assuming that the modification of an argument that then gets returned is part of the super's API? Either way might warrant a little comment.

asmtransformers/asmtransformers/models/embedder.py
```diff
@@ -0,0 +1,61 @@
...
+        convert_to_numpy=True,
+    ):
+        single_input = isinstance(sentences, str)
+        sentences = [sentences] if single_input else list(sentences)
+        normalize_embeddings = self.normalize_embeddings if normalize_embeddings is None else normalize_embeddings
+
+        embeddings = []
+        with torch.no_grad():
+            for start in range(0, len(sentences), batch_size):
+                batch = sentences[start : start + batch_size]
```
akaIDIOT: We're 3.13+, right? `itertools.batched()` is a thing now 🙏🏻 

Could also remove the need to `list(sentences)` above too, that reads like a performance drain somewhere down the line 😬

asmtransformers/tests/test_asmsentencebert_freeze.py
```diff
@@ -0,0 +1,82 @@
...
+                'do_lower_case': False,
+                'do_basic_tokenize': False,
+                'tokenize_chinese_chars': False,
+                'tokenizer_class': 'ARM64Tokenizer',
+                'model_max_length': 512,
+            }
+        )
+    )
+
+    torch.manual_seed(0)
```
akaIDIOT: This feels dangerous, is this 

- needed?
- the only way to achieve this?

Considering the type of project, we might at some point enable parallel tests, that might bite us a bit.

The fixture gets used a few times. If it's static, could it be session or module scoped?

asmtransformers/tests/test_embedder.py
```diff
@@ -43,34 +52,109 @@ def anchor():
...
+
+@pytest.mark.skipif(os.environ.get('CI') == 'true', reason="don't run this test on CI")
+def test_hf_embedding_model_loads_natively(anchor):
+    embedder = ASMEmbedder.from_pretrained('NetherlandsForensicInstitute/ARM64BERT-embedding')
+
+    embedding = embedder.encode(anchor)
+
+    assert np.isclose(embedding.sum(), -0.09272218)
+    assert np.isclose(embedding.min(), -0.10641833)
+    assert np.isclose(embedding.max(), 0.116405316)
```
akaIDIOT, line 170: 'Tis the same 🥳👌🏻

akaIDIOT (approved):

asmtransformers/asmtransformers/models/asmsentencebert.py
ranieri: My understanding is the typing.Mapping is immutable, and this one here is mutated. We could use MutableMapping ofc.

asmtransformers/tests/test_embedder.py
ranieri: 🥳

asmtransformers/tests/test_asmsentencebert_freeze.py
ranieri: Not necessary!

PR #43, opened 2026-05-01
process input as str by adding it to a list

### LINKED ISSUE #23: Tokenizer breaks on single-string input
ARM64Tokenizer mishandles a single input string. __call__() forwards straight to tokenize(), and tokenize() assumes texts is iterable over full samples; when given one string, it iterates character-by-character and json.loads() fails. That is a public API footgun because tokenizer callers normally expect both str and list[str] to work.

fleurvanl:


asmtransformers/asmtransformers/models/asmbert.py
```diff
@@ -200,6 +200,7 @@ def __init__(
 
     def tokenize(self, texts, split_special_tokens=False, **kwargs):
         encoded_inputs = []
+        texts = [texts] if isinstance(texts, str) else texts
```
ranieri, line 203: This works for me, but looking at transformers is see two distinct ways of doing this: the one above and checking the properties of  the first element.

I have no idea why they sometimes use one and sometimes the other.

https://github.com/huggingface/transformers/blob/7f6419e67de355ee173344c1bfd68cb60288e121/src/transformers/tokenization_python.py#L719

ranieri (approved):

akaIDIOT (approved):

PR #46, opened 2026-05-07
Add CONTRIBUTING.md

ranieri:


CONTRIBUTING.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Both Python projects currently require Python `>=3.13,<3.14`.
+
+Use the project-local PDM scripts:
+
+```bash
+pdm run check
+pdm run test
+pdm run all
```
akaIDIOT, line 47: The last one of these runs the first two again. What is the intention for the reader to do here?

akaIDIOT (approved):
Some things seem to be repeated a few times, though that might benefit both meatbags and clankers 🤷🏻

PR #47, opened 2026-05-07
Use sort when not DiGraph data entry in ARM64 preprocessor

AstroVasseL:
- Remove digraph dependency
-  use sort in arm64 preprocessor

### DIFF
--- a/asmtransformers/tests/test_arm64.py
+++ b/asmtransformers/tests/test_arm64.py
@@ -2,7 +2,6 @@ import importlib.resources
 import json
 
 import pytest
-from networkx import DiGraph
 
 from asmtransformers import arm64
 from asmtransformers.models.asmbert import ARM64Tokenizer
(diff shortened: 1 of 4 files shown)


asmtransformers/tests/test_arm64.py
```diff
@@ -51,10 +49,10 @@ def test_tokenize_single_block(tokenizer):
...
 
 def test_tokenize_branching_blocks(tokenizer):
-    graph = DiGraph()
-    # NB: nodes are in 'reverse order', tokenizer should (?) reorder these based on their node ids
-    graph.add_node(42, asm=['add x2,x2,#0x290', 'b.eq x2,0x0'])  # branch to offset 0
-    graph.add_node(0, asm=['sub x2,x2,#0x290', 'bl 0x2a'])  # branch to offset 42
+    graph = {
+        42: ['add x2,x2,#0x290', 'b.eq x2,0x0'],  # branch to offset 0
+        0: ['sub x2,x2,#0x290', 'bl 0x2a'],
+    }  # branch to offset 42
```
akaIDIOT, line 55: Comment is now outside the dict, should be on the same level as `# branch to offset 0`, right?

akaIDIOT (approved):

asmtransformers/tests/test_arm64.py
AstroVasseL: you are right

PR #49, opened 2026-05-08
Combined PR for ST inference-time elision and architecture support

ranieri:


ranieri:
"I" patched the compatibility wrapper in asmtransformers/models/st_compat.py.

  The key bit is in _normalize_encode_by_default():

```  get_model_kwargs = model.get_model_kwargs
  encode = model.encode

  def get_model_kwargs_with_architecture():
      return get_model_kwargs() + ['architecture']

  def encode_with_normalization_default(sentences, *args, normalize_embeddings=True, **kwargs):
      return encode(sentences, *args, normalize_embeddings=normalize_embeddings, **kwargs)

  model.get_model_kwargs = get_model_kwargs_with_architecture
  model.encode = encode_with_normalization_default
```

  SentenceTransformers only forwards selected kwargs down into the module/tokenizer path. By monkeypatching the returned ST compatibility instance’s get_model_kwargs() to include
  'architecture', model.encode(..., architecture='mips') is no longer dropped before tokenization.

  Then ASMTokenizer.tokenize() receives it here:

```
  def tokenize(self, texts, architecture='arm64', split_special_tokens=False, **kwargs):
      preprocessor = self.preprocessors[architecture]
```

  So architecture='mips' now reaches the tokenizer and raises the expected KeyError.

PR #51, opened 2026-05-11
Multi-arch mkvocab

ranieri:


asmtransformers/scripts/mkvocab.py
```diff
@@ -47,3 +58,7 @@
         for token in sorted(tokens):
             output.write(token)
             output.write('\n')
+
+
+if __name__ == '__main__':
+    mkvocab()
```
akaIDIOT: Might as well pull `sys.argv[1]` here instead above, right?

asmtransformers/scripts/mkvocab.py
```diff
@@ -5,37 +5,46 @@
...
-            operands.format_offset_log,
-        )
-    )
-    tokens = set()
 
-    for subset in dataset:
-        if SAMPLE_SIZE:
-            dataset[subset] = dataset[subset].select(range(SAMPLE_SIZE))
+    # Open without a vocab, the entire point is to make a new one.
+    tokenizer = asmbert.ASMTokenizer('/dev/null')
```
akaIDIOT, line 40: Considering reads from `/dev/null` will ... take a while, would `vocab=None` be clearer?

akaIDIOT (approved):

PR #52, opened 2026-05-11
Generalize preprocessors code

akaIDIOT:


asmtransformers/asmtransformers/preprocessors/arm64.py
```diff
@@ -0,0 +1,106 @@
...
+
+# a separator between operands; commas or whitespaces or a combination of both
+OPERAND_SEPARATOR = re.compile(r'[,\s]+')
+
+
+class ARM64Preprocessor(ASMPreprocessor):
+    branch_instructions = BRANCH_INSTRUCTIONS
+
+    def parse_operands(self, operands: str) -> Iterable[str]:
+        return parse_arm64_operands(operands)
```
ranieri: Why is the logic not implemented directly in this function? (see also other archs)

ranieri (approved):

asmtransformers/asmtransformers/preprocessors/arm64.py
akaIDIOT: Yeah, that's totally an option. There are tests that call these functions directly. I propose we use a separate PR for that?

On second thought, let's not go the separate PR, 'tis a silly place.

asmtransformers/asmtransformers/preprocessors/riscv.py
```diff
@@ -0,0 +1,74 @@
...
+    'bge',
+    'bgeu',
+    'bgez',
+    # jump
+    'j',
+    'jal',
+    'jalr',
+    # return
+    'ret',
+    'call',
```
akaIDIOT: Wondering whether these should count as branching instructions, they won't carry an offset operand, right? 🤔

asmtransformers/asmtransformers/preprocessors/x86.py
```diff
@@ -0,0 +1,154 @@
...
+    'jecxz',
+    'jrcxz',
+    # loop instructions (branch if cx != 0)
+    'loop',
+    'loope',
+    'loopne',
+)
+
+
+SIZE_QUALIFIERS = (
```
akaIDIOT: Inconsistent number of newlines between globals here.

asmtransformers/asmtransformers/preprocessors/riscv.py
akaIDIOT: Not for this PR, leaving this collection as is.

PR #53, opened 2026-05-12
fix risc-v negative immediates + compressed branch instructions

fleurvanl:


asmtransformers/asmtransformers/riscv.py
```diff
@@ -82,10 +94,9 @@ def preprocess(self, function_blocks: dict[int, list[str]]) -> list[str]:
 
                 tokens.append(instruction)
                 for operand in operands:
-                    if instruction in BRANCH_INSTRUCTIONS and is_offset(operand):
-                        # an operand to a branching instruction that is formatted as a hexadecimal number
-                        # this is interpreter as the offset to a basic block, and this tracked as such in path_offsets
-                        jump_target = int(operand[2:], base=16)
+                    if instruction in BRANCH_INSTRUCTIONS and (offset := is_offset(operand)):
+                        # can't slice at place 2 because negative hex values so therefore use value from is_offset regex
+                        jump_target = int(offset.group('value'), base=16)
```
akaIDIOT: The line comment refers to something the diff will remove. Won't that make the line comment incomprehensible without this diff? 

Also: all information in the previous comment is now gone 🤔

asmtransformers/tests/test_riscv.py
```diff
@@ -53,6 +53,30 @@ def test_jump_to_unknown_block(tokenizer):
     assert tokens.index('UNK_JUMP_ADDR') - tokens.index('bgt') == 1
 
 
+def test_compressed_jump():
+    tokens = riscv.RISCVPreprocessor(context_length=512).preprocess(
```
akaIDIOT: Test module has a thing like this available as a fixture already, right?


```suggestion
def test_compressed_jump(tokenizer):
    tokens = tokenizer.preprocess(
```

akaIDIOT (approved):

akaIDIOT:
Considering the PR title, this does nothing with immediate values, right?

asmtransformers/asmtransformers/riscv.py
```diff
@@ -73,6 +84,8 @@ def preprocess(self, function_blocks: dict[int, list[str]]) -> list[str]:
         # offsets we're collecting line up with the token offsets where those blocks actually start (see below)
         tokens = list(self.prefix_tokens)
 
+        function_blocks = dict(sorted(function_blocks.items()))
```
ranieri, line 87: This is in every preprocessor. Should we lift it to the caller?

akaIDIOT: See #52

PR #55, opened 2026-05-12
Align x86 operands parsing with other parsers

akaIDIOT:
Mostly: 

- searching for the next separator over searching for the end of an operand;
- using the same local var names as other parsers;
- adding some comments where non-simple things are happening;
- splitting some if/elif/else things into more cases inside match.

I'd have liked to do more of the last point, but the case guard would get too complex.

`x86.py` now has 100% test coverage too 🥳

### DIFF
--- a/asmtransformers/asmtransformers/preprocessors/x86.py
+++ b/asmtransformers/asmtransformers/preprocessors/x86.py
@@ -106,7 +106,7 @@ class X86Preprocessor(ASMPreprocessor):
                         assert rest.lstrip().startswith('ptr')
                         yield f'{token}_ptr'
                         # continue after "ptr"
-                        offset = end + rest.index('ptr') + 3
+                        offset = end + rest.index('ptr') + 3  # len('ptr')
                     else:
                         # strip optional segment-override suffix ":" from the token
                         # TODO: would it be informational to keep that suffix instead?
--- a/asmtransformers/asmtransformers/preprocessors/x86.py
+++ b/asmtransformers/asmtransformers/preprocessors/x86.py
@@ -68,10 +68,8 @@ SIZE_QUALIFIERS = {
 }
 
 
-# a separator between operands; commas, whitespaces, memory expressions or a combination of these
-OPERAND_SEPARATOR = re.compile(r'[,\s\[\]]+')
-# Matches one token that stops before commas, spaces, or memory-expression operators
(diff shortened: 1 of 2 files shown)


asmtransformers/asmtransformers/preprocessors/x86.py
```diff
@@ -85,70 +82,61 @@ def parse_operands(self, operands: str) -> Iterable[str]:
...
-                    if not prev or prev[-1] in ('+', '-', '*'):
-                        m = _OPERAND_TOKEN.match(expr, i + 1)
-                        if m:
-                            yield f'-{m.group().lower()}'
-                            i = m.end()
-                            continue
-                    yield '-'
-                    i += 1
+                    offset += 1
+                case '-' if not (preceding := expression[:offset].rstrip()) or preceding[-1] in '+-*':
```
AstroVasseL: would prefer this to be a const for readabiliy

asmtransformers/asmtransformers/preprocessors/x86.py
```diff
@@ -85,70 +82,61 @@ def parse_operands(self, operands: str) -> Iterable[str]:
...
-                            offset = m.end() + (len(operands[m.end() :]) - len(rest)) + 3
-                        else:
-                            yield lower
-                            offset = m.end()
+                        rest = operands[end:]
+                        # NB: a size qualifier *should* be followed by "ptr"
+                        assert rest.lstrip().startswith('ptr')
+                        yield f'{token}_ptr'
+                        # continue after "ptr"
+                        offset = end + rest.index('ptr') + 3
```
AstroVasseL: Would a line comment about the value '3' be neccesairy here?

AstroVasseL (commented):
Generally much nicer to read, and good that the preprocessor now is much more similar to other preprocessors. I like it a lot!

AstroVasseL (approved):

AstroVasseL:
especially the exponential increase of the Walrus makes me happy. Feel like Paul McCartney!

asmtransformers/asmtransformers/preprocessors/x86.py
akaIDIOT: Fair point, I shall add it (back, it was there before O:)).

akaIDIOT:
> especially the exponential increase of the Walrus makes me happy. Feel like Paul McCartney!

🦭

PR #58, opened 2026-05-19
Cluster ready pre-train

ranieri:


asmtransformers/docs/pretraining.md
```diff
@@ -0,0 +1,110 @@
...
+## Precision
+
+Use `--bf16` for CUDA bfloat16 mixed precision. This enables Hugging Face `TrainingArguments(bf16=True)`, so autocast
+uses bfloat16 where appropriate while preserving normal optimizer and checkpoint behavior.
+
+Use `--tf32` on Ampere, Hopper, or newer NVIDIA GPUs to allow TF32 matmul/cudnn kernels. The script enables PyTorch TF32
+backend flags and passes TF32 through to `TrainingArguments` when CUDA is available.
+
+Do not use these flags on older CUDA GPUs that do not support them. If a local CPU-only dry run or CI job needs to
+exercise CLI construction, `--bf16` and `--tf32` are accepted but not passed through to `TrainingArguments`.
```
akaIDIOT: Is that wise? If flags are explicitly passed on the command line, should the script maybe complain / crash instead of ignore these?

asmtransformers/scripts/pretrain.py
```diff
@@ -16,21 +17,90 @@ def timestamp():
...
+        dataloader_num_workers=dataloader_num_workers,
+        seed=seed,
+    )
+
+
+def load_eval_dataset(functions, eval_samples):
+    if 'test' not in functions:
+        return None
+
+    eval_tokenized = functions['test'].shuffle(seed=42)
```
akaIDIOT, line 87: Considering the nomenclature and all, should we expect there to be an `eval` section in the dataset? Or do test and eval mean the same here?

asmtransformers/scripts/pretrain.py
```diff
@@ -114,6 +188,10 @@ def pretrain(
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='ASM-Pretrain')
+    config_help = (
+        'the path of the model config used when initializing a new model; '
+        + 'defaults to packaged multilingual_asmbert'
+    )
```
akaIDIOT: Interesting place for this variable and interesting syntax in comparison with the rest of the code.

asmtransformers/scripts/pretrain.py
```diff
@@ -123,7 +201,7 @@ def pretrain(
     parser.add_argument(
         '--output-dir',
         type=str,
-        default=Path(__file__) / '../output',
+        default=Path(__file__).parent.parent / 'output',
```
akaIDIOT: Tempted to suggest that this gets no default. Would that hurt other things?

asmtransformers/scripts/slurm_pretrain.sh
akaIDIOT, line 1: Not familiar with most things in this file, to be honest :(

akaIDIOT (approved):
Not sure how to test this thoroughly, might just assume it works as intended 😅

asmtransformers/docs/pretraining.md
ranieri: Fails fast and loud now.

asmtransformers/scripts/pretrain.py
ranieri: I have no idea what happened there. Fixed.

asmtransformers/scripts/pretrain.py
ranieri: Agreed! It's a positional now.

asmtransformers/scripts/slurm_pretrain.sh
ranieri: The most salient parts are the `#SBATCH` comments, as that's the communication to the slurm job manager. Especially noteworthy is the --time: that sets how long the job is allowed to run. This impacts scheduling (need to wait for time for the entire job) and results (if the limit is reached your job is killed).

The rest is just a shell file.

PR #60, opened 2026-05-22
Finetune ready

fleurvanl:
should work combined with Ranieri's pr for a save-function

### DIFF
--- a/asmtransformers/scripts/ft-datasets.py
+++ b/asmtransformers/scripts/ft-datasets.py
@@ -1,15 +1,13 @@
 import argparse
-import json
 import os
 import random
 
-from datasets import Dataset
+from datasets import Dataset, DatasetDict
 from tqdm import tqdm
 
 
-def generate_train(data_folder, output_folder):
+def generate_train(data_folder):
     train_data_folder = os.path.join(data_folder, 'train')
-    train_output_folder = os.path.join(output_folder, 'train')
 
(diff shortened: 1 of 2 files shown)


asmtransformers/scripts/ft-datasets.py
```diff
@@ -93,6 +94,8 @@ def add_random(example):
 def main(data_folder, output_folder):
     generate_train(data_folder, output_folder)
     generate_eval(data_folder, output_folder)
+    with open(os.path.join(output_folder, 'dataset_dict.json'), 'w') as f:
```
ranieri: This should probably be achieved by actually building a `DatasetDict` object and saving that.

ranieri (approved):

asmtransformers/scripts/ft-datasets.py
```diff
@@ -87,12 +83,14 @@ def add_random(example):
...
+    return trip
 
 
 def main(data_folder, output_folder):
-    generate_train(data_folder, output_folder)
-    generate_eval(data_folder, output_folder)
+    train = generate_train(data_folder)
+    test = generate_eval(data_folder, output_folder)
+    dataset_dict = DatasetDict({'train': train, 'test': test})
+    dataset_dict.save_to_disk(output_folder)
```
ranieri, line 93: 👍

PR #62, opened 2026-05-22
Avoid dir name race, harden logging in cluster mode

ranieri:
Derive output folder from SLURM job id if present to avoid an issue where different nodes use different timestamps for the model folder name.


AstroVasseL (approved):

ranieri:
I'm holding off on merging this for now, still feels like it does not resolve the root cause.

ranieri:
In hindsight, the fix was a little too minimal. I added a --run-id parameter that allows the user to specify the a custom value for non-slurm cluster runs and other special occasions.

PR #63, opened 2026-05-26
Add save and load functionality to ASMTransformerModule

ranieri:


ranieri:
The ST model needs to be able to save and load itself (as an ST model) for checkpointing.

fleurvanl (approved):

PR #65, opened 2026-05-29
use map-function to parallelize extract tokens

fleurvanl:
keep in mind that we use a map function in a way that it works like a reduce function, see https://colab.research.google.com/drive/1jCLv31Y4cDfqD0lhO0AnqEv3Or-LLvWe?usp=sharing#scrollTo=HI7-FPO4QJaI (final function)

### DIFF
--- a/asmtransformers/tests/test_mktokenizer.py
+++ b/asmtransformers/tests/test_mktokenizer.py
@@ -4,6 +4,7 @@ from datasets import Dataset
 from asmtransformers.models import asmbert
 from scripts.mktokenizer import extract_tokens_map
 
+
 @pytest.fixture
 def dataset():
     return Dataset.from_dict(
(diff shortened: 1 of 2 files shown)


ranieri:
Verified on `/mnt/projects/fbda/Semantic_Code_Search/Data/DEB-bigger-selection/function-dataset`. Set runs in about 1 minute, using 8 cores.

asmtransformers/tests/test_mktokenizer.py
```diff
@@ -0,0 +1,39 @@
...
+                'arm64',
+                'i386',
+                'i386',
+                'riscv64',
+            ],
+        }
+    )
+
+
+def test_extract_tokens_map(dataset):
```
ranieri, line 34: Thanks for adding the test!

ranieri (approved):

PR #67, opened 2026-06-04
let tokenize_dataset also accept of type datasets.Dataset

AstroVasseL:
Small type check in tokenize_dataset that also accepts type of datasets.Dataset as input to make a tokenized dataset when a train/test split has not happened yet.

### DIFF
--- a/asmtransformers/scripts/tokenize_dataset.py
+++ b/asmtransformers/scripts/tokenize_dataset.py
@@ -25,7 +25,7 @@ if __name__ == '__main__':
     dataset = datasets.load_from_disk(data_in)
 
     # let the tokenizer preprocess data from data_in, write the result to data_out
-    for subset in dataset:
-        dataset[subset] = tokenize(tokenizer, dataset[subset])
+    # for subset in dataset:
+    dataset = tokenize(tokenizer, dataset)
 
     dataset.save_to_disk(data_out)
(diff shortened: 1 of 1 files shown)


asmtransformers/scripts/tokenize_dataset.py
```diff
@@ -25,7 +25,11 @@ def do_tokenize(function):
...
 
     # let the tokenizer preprocess data from data_in, write the result to data_out
-    for subset in dataset:
-        dataset[subset] = tokenize(tokenizer, dataset[subset])
+
+    if isinstance(dataset, datasets.Dataset):  # datasets.load_from_disk either a Dataset or DatasetDict type
+        dataset = tokenize(tokenizer, dataset)
+    else:
+        for subset in dataset:
+            dataset[subset] = tokenize(tokenizer, dataset[subset])
```
akaIDIOT, line 34: Is the second case a mapping, perchance? We *could* swap that into a `case {}:` O:)

akaIDIOT (approved):

asmtransformers/scripts/tokenize_dataset.py
ranieri: It is. The problem however is that the `Dataset` itself is also a mapping, with the columns as items.

ranieri (approved):

PR #69, opened 2026-06-11
add quality score script

AstroVasseL:


asmtransformers/scripts/quality_score.py
```diff
@@ -0,0 +1,59 @@
...
+
+if __name__ == '__main__':
+    # With mktokenizer and tokenize_dataset.py
+    tokenized_dataset = datasets.load_from_disk(sys.argv[1])
+
+    # Need two passes of `map`` because to normalize we have to know min and max of values
+    tokenized_dataset = tokenized_dataset.map(map_tokens_to_df)
+    # In this pass we determine quality_score
+    scored_dataset = tokenized_dataset.map(make_scorer(tokenized_dataset))
+    scored_dataset.save_to_disk(OUTPUT)
```
akaIDIOT: Would it make sense to take this from `sys.argv`?

asmtransformers/scripts/quality_score.py
```diff
@@ -0,0 +1,69 @@
...
+    return add_score
+
+
+if __name__ == '__main__':
+    # With mktokenizer and tokenize_dataset.py
+    dataset_path, tokenizer_path = sys.argv[1:]
+    tokenized_dataset = datasets.load_from_disk(sys.argv[1])
+
+    # Need two passes of `map`` because to normalize we have to know min and max of values
+    tokenized_dataset = tokenized_dataset.map(map_tokens_to_dataset, fn_kwargs={'tokenizer_path': tokenizer_path})
```
akaIDIOT: Does it make sense to load the tokenizer here and pass the tokenizer as an object? Seems silly to have one pre-loaded from file and a file name, while the latter can also be loaded before calling `dataset.map()`.

asmtransformers/scripts/quality_score.py
```diff
@@ -0,0 +1,68 @@
...
+
+        return example
+
+    return add_score
+
+
+if __name__ == '__main__':
+    # Make sure tokenized dataset has been made With mktokenizer and tokenize_dataset.py
+    dataset_path, tokenizer_path = sys.argv[1:]
+    tokenized_dataset = datasets.load_from_disk(sys.argv[1])
```
akaIDIOT: `sys.argv[1]` is used twice now, the second occurrence can be `dataset_path`, I take it?


```suggestion
    dataset_path, tokenizer_path = sys.argv[1:]
    tokenized_dataset = datasets.load_from_disk(dataset_path)
```

asmtransformers/scripts/quality_score.py
```diff
@@ -0,0 +1,69 @@
...
+
+def map_tokens_to_dataset(dataset, tokenizer):
+    jtp_out_of_range_token = tokenizer['model']['vocab']['JUMP_ADDR_EXCEEDED']
+    jtp_unknown = tokenizer['model']['vocab']['UNK_JUMP_ADDR']
+    minlength = max(jtp_out_of_range_token, jtp_unknown) + 1
+    token = np.asarray(dataset['input_ids']).ravel()
+    token_to_id = {t['content']: t['id'] for t in tokenizer['added_tokens']}
+    padding_token = token_to_id['[SEP]']
+    token = token[token != padding_token]
+    bincount = np.bincount(token, minlength=minlength)
```
akaIDIOT, line 16: I cannot get to grips with numpy logic, but as discussed offline, I think i get it.

akaIDIOT (approved):

PR #74, opened 2026-06-16
Remove redundant shuffle

ranieri:


asmtransformers/scripts/pretrain.py
```diff
@@ -226,7 +226,7 @@ def pretrain(
 
     # Load the training and evaluation datasets
     functions = load_from_disk(data)
-    train_tokenized = functions['train'].shuffle(seed=42)
+    train_tokenized = functions['train']
```
fleurvanl, line 229: Why don't we move the assert here from the test? Right now if pretrain.py changes our test still passes, regardless of whether this data is sampled randomly. With the assert here we get at least an error telling us if the data is not shuffled

fleurvanl (approved):

PR #75, opened 2026-06-16
HOTFIX: don't accidentally iterate the whole dataset

ranieri:


asmtransformers/scripts/pretrain.py
```diff
@@ -226,7 +226,7 @@ def pretrain(
 
     # Load the training and evaluation datasets
     functions = load_from_disk(data)
-    if 'train' in functions:
+    if isinstance(functions, DatasetDict) and 'train' in functions:
```
akaIDIOT, line 229: Ghe, nice catch. How about we don't check many many millions of datapoints 🤡

akaIDIOT (approved):

PR #77, opened 2026-06-17
Add annealing phase to pretraining

AstroVasseL:


asmtransformers/docs/pretraining.md
```diff
@@ -130,6 +72,46 @@ nodes * gpus_per_node * batch_size * gradient_accumulation_steps
 For example, 2 nodes with 8 GPUs each, `BATCH_SIZE=16`, and `GRADIENT_ACCUMULATION_STEPS=4` gives a global batch size
 of 1024 sequences.
 
+## Quality Annealing (Final Phase)
+
+The last phase of pretraining oversamples high-quality functions so the model finishes on a corpus skewed toward
+clean control-flow graphs. This is a separate, short run that continues from the phase-1 model.
```
fleurvanl: what does it mean "clean control-flow graphs"? are those the ones that fit in vocab?

asmtransformers/scripts/quality_score.py
fleurvanl: I know this isn't part of your pr, but I find it really annoying that I have to point to the specific tokenizer_config.json instead of the folder. Maybe allign this with the argparse in other  files? I think ft-datasets.py was exactly like this so you can probably copy the argparse from there (and then put a comment that you want the tokenizer_config.json specifically)

fleurvanl (approved):

asmtransformers/scripts/evaluation.py
```diff
@@ -143,6 +143,9 @@ def generate_test_pools(data_folder, pool_size, static_pool):
     return
     Generator yielding triplets: anchor, positive, numpy.array(negative_embeddings * POOL_SIZE)"""
     dataset = datasets.load_from_disk(data_folder)  # .select(range(11000, 45000))
+    if architecture:
+        print(f'Selecting examples with architecture=={architecture}')
```
akaIDIOT, line 147: Print needed here? Not much of the code does debug-prints, this seems like a debugging leftover?

asmtransformers/scripts/oversample_quality.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+            }
+        )
+
+    raise TypeError(f'unsupported dataset type {type(scored).__name__}')
+
+
+def build_arg_parser():
+    parser = argparse.ArgumentParser(
+        description='Build a quality-oversampled training set for the final pretraining (annealing) phase'
```
akaIDIOT, line 76: Help text mixes quality and scores; *implying* the same thing. Should we just use score to be consistent?

asmtransformers/scripts/oversample_quality.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    parser.add_argument('input', help='path to the scored dataset saved by scripts/quality_score.py')
+    parser.add_argument('output', help='path to write the oversampled DatasetDict (save_to_disk)')
+    parser.add_argument(
+        '--top-fraction',
+        type=float,
+        default=0.25,
+        help='fraction of the corpus treated as high quality (default: top 25%%)',
```
akaIDIOT, line 84: Why Why the the double double percent percent?? 🤡

asmtransformers/scripts/oversample_quality.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    parser.add_argument(
+        '--repeats',
+        type=int,
+        default=3,
+        help='number of extra copies of each high-quality function (default: 3)',
+    )
+    parser.add_argument('--score-column', default='score', help="name of the quality-score column (default: 'score')")
+    parser.add_argument('--seed', type=int, default=42, help='shuffle seed for the oversampled training rows')
```
akaIDIOT, line 93: This will set a seed regardless of command line arguments; would no seed / actual random be a better default?

asmtransformers/scripts/quality_score.py
```diff
@@ -50,17 +68,30 @@ def add_score(example):
...
 
-if __name__ == '__main__':
-    # Make sure tokenized dataset has been made With mktokenizer and tokenize_dataset.py
-    dataset_path, tokenizer_path, output_path = sys.argv[1:]
-    with open(tokenizer_path) as f:
+def get_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument('-d', '--data-folder', type=str, required=True, help='folder with data')
+    parser.add_argument('-t', '--tokenizer', type=str, required=True, help='path to tokenizer.json')
+    parser.add_argument('-o', '--output-folder', type=str, required=True, help='folder with data')
```
akaIDIOT: Other script uses `input` and `output` positional arguments. We should use the same way of supplying inputs and outputs throughout all the scripts, I'd say. Inconsistency is painful here.

asmtransformers/scripts/tokenize_dataset.py
```diff
@@ -17,19 +17,41 @@ def do_tokenize(function):
...
-    dataset.save_to_disk(data_out)
+    dataset.save_to_disk(output_folder)
+
+
+def get_parser():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--tokenizer', type=str, required=True, help='folder with tokenizer')
+    parser.add_argument('--input-data', type=str, required=True, help='data to be used for training')
+    parser.add_argument('--output-folder', type=str, required=True, help='folder to leave the tokenized data')
```
akaIDIOT, line 43: Same inconsistent argument handling handler here.

akaIDIOT (approved):

asmtransformers/scripts/oversample_quality.py
fleurvanl: if I run the --help it comes out as "(default: top 25%)" (so single %-sign)

asmtransformers/scripts/oversample_quality.py
fleurvanl: Do you mean "build a score-oversampled training set"? because to me that feels less clear

asmtransformers/scripts/oversample_quality.py
akaIDIOT: Maybe, but the value a user would have to put there, is `0.25`; supplying `--top-fraction 25%` would crash, no?

asmtransformers/scripts/oversample_quality.py
akaIDIOT: Hmm, the `--help` output also uses terms like "quality-scored". I guess it's fine, but the two terms mingling was confusing when I read it the first time 🤔

PR #79, opened 2026-06-23
Update label generation to use `file_name` and `function_name` fields

ranieri:


akaIDIOT (approved):
That's my bad, completely forgot to align the output with the way things were a year ago 🙈

PR #80, opened 2026-06-23
Eval by architecture

ranieri:


asmtransformers/scripts/evaluation.py
```diff
@@ -22,7 +22,7 @@ def add_label(example):
 
     :param example: row in dataset
     :returns example with an extra column containing the label"""
-    label = example['bin_name'] + '/' + example['func_name']
+    label = example['file_name'] + '/' + example['function_name']
```
akaIDIOT, line 25: This line is the same as the other PR, right? Should drop out of it if it merges before this one?

akaIDIOT (approved):

PR #81, opened 2026-06-26
Add support for repeated evals of a static pool

ranieri:


asmtransformers/scripts/evaluation.py
```diff
@@ -49,8 +41,10 @@ def generate_neg_pool(pool_size, dataset, anchor_labels, anchor_cfgs, pos_cfgs):
     :return: a numpy array containing the embeddings of the items in the pool
     """
     neg_embeddings = []
-    while len(neg_embeddings) < pool_size:
-        neg = generate_single_neg(dataset)
+    candidate_indices = list(range(len(dataset)))
+    random.shuffle(candidate_indices)
+    for index in candidate_indices:
```
akaIDIOT, line 47: This seems like a fairly wasteful way of randomizing selections, `random.choices(range(len(dataset))` won't populate a `list`, or `dataset.shuffle` might help?

asmtransformers/scripts/evaluation.py
```diff
@@ -151,8 +146,21 @@ def generate_test_pools(data_folder, pool_size, static_pool, architecture=None):
...
+
+def generate_test_pools(data_folder, pool_size, static_pool, architecture=None):
+    """order data in such a way that we can make triplets consisting of an anchor, a positive and pool_size * negative
+    examples, then call generate_triplets() to generate said triplets
+    :param data_folder: Path to data
+    :param pool_size: number of negative items to compare with
+    :param static_pool: use the same pool of negatives for every pos/anchor pair (faster)
+
+    return
+    Generator yielding triplets: anchor, positive, numpy.array(negative_embeddings * POOL_SIZE)"""
```
akaIDIOT: Docstring formatting reads confusing, what's the intention here?

asmtransformers/scripts/evaluation.py
```diff
@@ -197,32 +207,87 @@ def calculate_all(test_pools, output_path, output_file):
...
+
+
+def run_tests(data_folder, output_path, pool_size, static_pool, architecture, seed, repeats=1):
+    if repeats < 1:
+        raise ValueError('repeats must be at least 1')
+    if repeats > 1 and not static_pool:
+        raise ValueError('repeats greater than 1 are only supported with --static-pool')
 
+    if seed is not None:
+        random.seed(seed)
```
akaIDIOT: Do we really need to seed the global random? If we'll ever edit this file in seemingly innocent ways, the same seed will never generate the same result again. Can we seed things more local to where they're needed? And leave the global random state alone? 😬

asmtransformers/scripts/evaluation.py
```diff
@@ -197,32 +207,87 @@ def calculate_all(test_pools, output_path, output_file):
...
+
     model_name = data_folder.split('/')[-1]
     output_file = f'{model_name}-{pool_size}-{static_pool}-{timestamp()}'
-    calculate_all(test_pools, output_path, output_file)
+    repeat_seeds = [None] * repeats if seed is None else [seed + repeat + 1 for repeat in range(repeats)]
+    aggregate_rows = []
+
+    for repeat, repeat_seed in enumerate(repeat_seeds):
+        if repeat_seed is not None:
+            random.seed(repeat_seed)
```
akaIDIOT: Same here; is it possible to leave the global random alone?

akaIDIOT (approved):

asmtransformers/scripts/evaluation.py
ranieri: Gone

asmtransformers/scripts/evaluation.py
ranieri: `random.choices()` samples with replacement (see https://docs.python.org/3/library/random.html#random.choices).

We have a variable number of candidates that we accept or reject until we have pool_size values. Therefore we can't start with pool_size samples, and we can't predict how many we will need. We can use `random.sample(range(n), n)`, but this is not materially different from in-place shuffle as we're still left with a list of length n in memory.

The cleanest way to do this is to sample a new random index every loop and keep track of the visited samples to avoid repeating. This is more involved and - without additional care - behaves badly when there aren't enough candidates: it becomes very slow as most samples fail and then freezes.

A final option is to use a function that visits all N elements in random order, a pseudo random permutation (PRP, https://en.wikipedia.org/wiki/Pseudorandom_permutation). A possible option is a block-cipher based  Black-Rogaway construction (https://en.wikipedia.org/wiki/Format-preserving_encryption). While fun, this is not really a thing we want to spend time on right now.

On balance, this is probable fine as is.

PR #84, opened 2026-06-30
normalize argparse in /scripts

fleurvanl:
input & output args are now  always positional arguments (thus no --something or -s)
sometimes model/tokenizer are also positional, if they were required anyways

### DIFF
--- a/asmtransformers/scripts/finetune.py
+++ b/asmtransformers/scripts/finetune.py
@@ -126,7 +126,7 @@ def main(data_folder, model, batch_size):
 
 def get_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser()
-    parser.add_argument('data_folder', type=str, help='folder with data')
+    parser.add_argument('data_folder', type=str, help='folder with input data')
 
     parser.add_argument('model', type=str, help='The name of the model used for finetuning')
 
--- a/asmtransformers/scripts/finetune.py
+++ b/asmtransformers/scripts/finetune.py
@@ -126,9 +126,9 @@ def main(data_folder, model, batch_size):
 
 def get_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser()
-    parser.add_argument('-d', '--data-folder', type=str, required=True, help='folder with data')
+    parser.add_argument('data_folder', type=str, help='folder with data')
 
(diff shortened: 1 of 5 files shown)


asmtransformers/scripts/inference.py
```diff
@@ -19,10 +19,10 @@ def main(data_folder, output_folder, model_path):
 
 def get_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(add_help=False)
-    parser.add_argument('-d', '--data-folder', type=str, required=True, help='folder with data')
-    parser.add_argument('-o', '--output-folder', type=str, required=True, help='folder with data')
+    parser.add_argument('data_folder', type=str, help='folder with data')
+    parser.add_argument('output_folder', type=str, help='folder with data')
```
akaIDIOT: Help texts are not very distinguishing here, I'm assuming the order is `input, output`?

asmtransformers/scripts/ft-datasets.py
```diff
@@ -95,8 +95,8 @@ def main(data_folder, output_folder):
 
 def get_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(add_help=False)
-    parser.add_argument('-d', '--data-folder', type=str, required=True, help='folder with data')
-    parser.add_argument('-o', '--output-folder', type=str, required=True, help='folder with data')
+    parser.add_argument('data_folder', type=str, help='folder with data')
+    parser.add_argument('output_folder', type=str, help='folder with data')
```
akaIDIOT: Help texts are not very distinguishing here, I'm assuming the order is `input, output`?

asmtransformers/scripts/finetune.py
```diff
@@ -126,9 +126,9 @@ def train_callback(score, epoch, steps):
 
 def get_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser()
-    parser.add_argument('-d', '--data-folder', type=str, required=True, help='folder with data')
+    parser.add_argument('data_folder', type=str, help='folder with data')
```
akaIDIOT: *input* data?

akaIDIOT (approved):
Nice to normalize these things (I think, at least :D)

Looking at the file list, we're also not very consistent with our file names; do scripts use dashes or underscores? 🤡 (maybe not something you need to fix, we all just need to look at our previous work a bit more when making something new, I guess 🙈)

PR #85, opened 2026-06-30
Add PostgreSQL database interface

akaIDIOT:
Note that this does not *use* the new implementation in the API bit, it's a second implementation of `Database`, that's not really adhering to the protocol as it's all async. We can fix this in subsequent PR's where the sqlite implementation is removed.

### DIFF
--- a/.github/workflows/citatio.yml
+++ b/.github/workflows/citatio.yml
@@ -34,7 +34,7 @@ jobs:
         ports:
           - 5432:5432
         env:
-          POSTGRES_PASSWORD: wieditleestisgek
+          POSTGRES_PASSWORD: tee4ieCh1jeerahkie3u
         options: >-
           --health-cmd pg_isready
           --health-interval 10s
@@ -53,4 +53,4 @@ jobs:
         env:
           POSTGRES_HOST: localhost
           POSTGRES_PORT: 5432
-          POSTGRES_PASSWORD: wieditleestisgek
+          POSTGRES_PASSWORD: tee4ieCh1jeerahkie3u
--- a/.github/workflows/citatio.yml
+++ b/.github/workflows/citatio.yml
@@ -25,18 +25,32 @@ jobs:
   test:
     runs-on: ubuntu-latest
     needs: check
-    strategy:
-      matrix:
-        python-version: ['3.13']
     defaults:
       run:
         working-directory: citatio/
+    services:
+      pgvector:
+        image: pgvector/pgvector:pg18
+        ports:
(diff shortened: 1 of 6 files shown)


.github/workflows/citatio.yml
```diff
@@ -25,18 +25,32 @@ jobs:
...
     defaults:
       run:
         working-directory: citatio/
+    services:
+      pgvector:
+        image: pgvector/pgvector:pg18
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_PASSWORD: wieditleestisgek
```
ranieri: Ehhhhhhhh....

citatio/citatio/db.py
```diff
@@ -106,11 +108,65 @@ def search_function(self, embedding, top_n=25):
...
+        self.connection = connection
 
-    def add_function(self, name, cfg, embedding, binary_name, binary_sha256, model_identifier=None):
-        raise NotImplementedError
+    @classmethod
+    async def connect(cls, **kwargs):
+        connection = await asyncpg.connect(**kwargs)
+        await connection.execute('CREATE EXTENSION IF NOT EXISTS vector')
+        await register_vector(connection)
+        await connection.execute(resources.read_text('citatio', 'schema-postgresql.sql'))
```
drxzcl, line 119: We run the init script every time, and the init script is idempotent because of the IF NOT EXIST bits. What's the plan for schema upgrade?

drxzcl (approved):

ranieri (approved):

citatio/citatio/db.py
akaIDIOT: Blup, comment got deleted when I pressed resolve...

Correct, we're not currently dealing with schema changes (sqlite implementation wasn't either, btw). Logged this as a separate issue, see #93

.github/workflows/citatio.yml
akaIDIOT: Goooood pooooint... 😬 Not sure what to put here really, [the official docs](https://docs.github.com/en/actions/tutorials/use-containerized-services/create-postgresql-service-containers) simply put "postgres" as the connection password, which to me was no better. 

I *believe* it's a local container only usable from the running action, so this leaking should maybe not matter too much? Going from the docs, I'm not sure what else to put here, unless we think this value isn't very professional? 

(I mean, it isn't, but is that an issue? 😇)

akaIDIOT: Replaced with a generated value.

PR #87, opened 2026-07-02
Let FastAPI app use postgresql database

akaIDIOT:
- [x] contains #85, which should be merged first

### DIFF
--- a/citatio/citatio/api.py
+++ b/citatio/citatio/api.py
@@ -16,10 +16,10 @@ DEFAULT_MODEL = 'NetherlandsForensicInstitute/ARM64BERT-embedding'
 @asynccontextmanager
 async def lifespan(app: FastAPI):
     config = confidence.load_name('citatio')
-    if not config.model or config.database:
-        raise ValueError(f'missing model or database configuration: {config}')
+    if not config.database:
+        raise ValueError(f'missing database configuration, available: {config}')
 
-    app.state.model = ASMEmbedder.from_pretrained(config.model)
+    app.state.model = ASMEmbedder.from_pretrained(config.model or DEFAULT_MODEL)
 
     async with await PostgreSQLDatabase.connect(**config.database) as database:
         app.state.database = database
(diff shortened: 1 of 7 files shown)


citatio/citatio/api.py
```diff
@@ -15,26 +15,22 @@
...
 
-    if database == ':memory:':
-        # TODO: log warning that we're running in in-memory database
-        pass
+    app.state.model = ASMEmbedder.from_pretrained(config.model or DEFAULT_MODEL)
 
-    app.state.model = ASMEmbedder.from_pretrained(model)
-
-    with SQLiteDatabase.from_name(database) as database:
+    async with await PostgreSQLDatabase.connect(**config.database) as database:
```
ranieri, line 24: Let's have a quick "how to use async stuff" session next week. This looks intimidating if you're not used to it.

ranieri:
Is there an easy way to ignore the changes that are also in the previous PR?

ranieri (approved):

citatio/citatio/api.py
akaIDIOT: It's intimidating for me too; let's have a quicky about this soon, though I'm not overly confident about these things either.

akaIDIOT:
> Is there an easy way to ignore the changes that are also in the previous PR?

The diff doesn't even update when the other PR gets merged. I really want a better way to build sequences of PRs… 🤔 Rebase + force-push incoming…

citatio/citatio/db.py
```diff
@@ -147,7 +153,7 @@ async def add_function(self, name, cfg, embedding, binary_name, binary_sha256, m
     async def _search_near(self, embedding, top_n):
         return await self.connection.fetch(
             """
-            SELECT label, (2 - (embedding <-> $1)) / 2 AS similarity, binary_name, binary_sha256
+            SELECT label, (2 - (embedding <=> $1)) / 2 AS similarity, binary_name, binary_sha256
```
akaIDIOT, line 156: Not really related, `<->` was taken from an example, it turns out to be l2-distance, `<=>` is cosine distance. As that's the distance metric we've been using, I figured this should be changed along with it.

PR #90, opened 2026-07-03
Make sqlite interface async, parametrize both database and api tests to use both sqlite and postgresql in tests

### LINKED ISSUE #2: Enable using PostgreSQL as the database
SQLite has been great for development, production use cases should really support PostgreSQL. Both API routes in citatio are currently synchronous / blocking due to observed issues with SQLite, these issues should be gone with PostgreSQL too.

akaIDIOT:
Requires PR's to be merged:

- [x] #85 
- [x] #87 

Closes #2


citatio/citatio/api.py
```diff
@@ -15,26 +15,29 @@
...
-    database = os.environ.get('CITATIO_SQLITE_DATABASE', ':memory:')
-
-    if database == ':memory:':
-        # TODO: log warning that we're running in in-memory database
-        pass
-
-    app.state.model = ASMEmbedder.from_pretrained(model)
-
-    with SQLiteDatabase.from_name(database) as database:
+    config = confidence.load_name('citatio')
```
akaIDIOT, line 18: No specifying a format here implies we'll be using YAML format and `confidence.load_name` will attempt to find `citatio.yaml` files. Tests are using environment variables to configure the lifespan into using either sqlite or postgresql, but we might at some point tell potential users they *could* use a file (this might be nice for postgresql, that requires multiple values to be configured when not using a "dsn"). 

Concrete question: should we declare `formats.TOML` here, or explicitly put `formats.YAML` here?

citatio/citatio/api.py
```diff
@@ -15,26 +15,29 @@
...
+
+    match config:
+        case {'database.sqlite': name}:
```
akaIDIOT, line 21: TIL: A `confidence.Configuration` object will match both `{'database': {'sqlite': name}}` and `{'database.sqlite': name}` here 👏🏻

citatio/citatio/api.py
```diff
@@ -15,26 +15,29 @@
...
         app.state.database = database
-
         yield
 
 
 app = FastAPI(lifespan=lifespan)
 
 
 @app.post('/api/v1/add')
-async def add_function(  # NB: function body isn't actually async, forcing it to run blocking on the event loop
```
akaIDIOT: Not async comments removed, as the `database` object is actually using `async` methods, though ... the sqlite implementation will still block the event loop 😅

citatio/tests/conftest.py
```diff
@@ -17,3 +23,116 @@ def functions():
 def embeddings(functions):
     arrays = np.load(TEST_ROOT / 'functions.npy', allow_pickle=False)
     return {function['name']: embedding for function, embedding in zip(functions, arrays, strict=True)}
+
+
+@pytest.fixture(scope='session')
+def local_pgvector_container():
```
akaIDIOT, line 29: This is deliberately made a session-scoped fixture to keep it around for the entire session. Though never directly used as a function-parameter fixture, it's requested as through `request.getfixturevalue()` below.

citatio/tests/conftest.py
akaIDIOT, line 1: This file juggles environment variables quite a bit, supporting 3 different things:

- sqlite database api;
- postgresql database api taken from a container object;
- postgresql database api taken from environment variables set from github actions.

It reads like a fever dream now, I'll see if I can improve it.

akaIDIOT: Improvements were made 🫡

citatio/tests/conftest.py
```diff
@@ -26,3 +23,91 @@ def functions():
...
+@pytest.fixture
+async def database_env(monkeypatch, database_config):
+    match database_config:
+        case {'database.sqlite': ':memory:'}:
+            # NB: add quotes to avoid the configuration's format misparsing :memory:
+            monkeypatch.setenv('CITATIO_DATABASE_SQLITE', '":memory:"')
+            yield
+        case {'database.sqlite': fname}:
+            monkeypatch.setenv('CITATIO_DATABASE_SQLITE', fname)
+            yield
```
akaIDIOT, line 101: Figured I'd include this for completeness, but it's never used as using an in-memory sqlite database is a lot simpler. Remove or retain this?

citatio/tests/test_db.py
akaIDIOT, line 1: All tests in this file are identical from what was already in `test_db_postgresql.py` and `test_db_sqlite.py`, but github decides to show it as new 🤷🏻

citatio/tests/conftest.py
```diff
@@ -26,3 +23,91 @@ def functions():
...
+                'user': container.username,
+                'password': container.password,
+                'database': container.dbname,
+            }
+
+
+@pytest.fixture(
+    params=[
+        pytest.param('sqlite', marks=pytest.mark.sqlite),
+        pytest.param('postgresql', marks=pytest.mark.postgresql),
```
akaIDIOT, line 63: Actual selection + marker magic is here:

- the 2 params automagically parametrizes any test that uses this fixture and gets run once with each;
- the marks on the params automagically mark any test instance with the applicable mark, making them selectable through `pytest run` (see `pyproject.toml` run aliases).

citatio/pyproject.toml
ranieri: Shouldn't this be a test dependency?

citatio/tests/conftest.py
ranieri: It took me at least three explanations from the author, in addition to this comment, to get this. Maybe it should be a more didactically inclined line comment.

citatio/tests/conftest.py
ranieri: Other option: retain bottom one, set env to `:memory:`

ranieri:
Is the readme bit relating to default `:memory:` DB still true?

citatio/citatio/db.py
```diff
@@ -66,10 +65,10 @@ def _insert_or_get_function(self, cursor, cfg, embedding):
 
         return function_id
 
-    def add_function(self, name, cfg, embedding, binary_name, binary_sha256, model_identifier=None):
+    async def add_function(self, name, cfg, embedding, binary_name, binary_sha256, model_identifier=None):
         with self.connection:
             cursor = self.connection.cursor()
-            function_id = self._insert_or_get_function(cursor, cfg, embedding)
+            function_id = await self._insert_or_get_function(cursor, cfg, embedding)
```
ranieri, line 71: Should we wrap `_insert_or_get_function` and `INSERT INTO labels...` with a transaction?

Atomicity was always a problem, but it feels worse now because we have potentially a real DB server with multiple clients and "await" just reads as "things can happen here".

citatio/citatio/api.py
ranieri: I'd leave this as-is for this PR, since we don't actually have any files.

In the future, we can pick a format and document it. I have a slight pref for TOML (because Norway), but that's not a choice we need to make now.

ranieri (approved):

citatio/pyproject.toml
akaIDIOT: Whoops, it should!

citatio/tests/conftest.py
akaIDIOT: Good idea, quoted `fname` to avoid the same issue as above.

citatio/tests/conftest.py
akaIDIOT: *Donning the didactician's wise hat*

akaIDIOT: As discussed offline, sorted!

citatio/citatio/db.py
akaIDIOT: Made it all simpler:

- insert-or-get is now a single query (using conflict resolution to trick the `RETURNING id` into working);
- few functions that were separated for the sake of being able to use async are now inlined into a single statement;
- wrapping the dual insert statements into a transaction as suggested 🙏🏻

akaIDIOT:
> Is the readme bit relating to default `:memory:` DB still true?

Good catch, updated the README to reflect the new options.

PR #92, opened 2026-07-08
Ensure finetune is random shuffled, finetune using input_id.

ranieri:


asmtransformers/asmtransformers/datasets/sentencelabel.py
```diff
@@ -3,34 +3,12 @@
 from operator import itemgetter
 
 from datasets import Dataset
-from sentence_transformers import InputExample
 from torch.utils.data import IterableDataset
 from tqdm import tqdm
 
 
 class LazySentenceLabelDataset(IterableDataset):
-    """A lazy-loading version of sentence_transformers.datasets.SentenceLabelDataset designed for use with Hugging
```
fleurvanl, line 12: Where did the extensive docstring go? Is al this information no longer correct?

asmtransformers/scripts/finetune.py
```diff
@@ -87,7 +127,7 @@ def eval_triplets(functions):
             yield InputExample(texts=[example['anchor'], example['pos'], example['neg']])
 
     test_functions = functions['test']
-    dev_evaluator = TripletEvaluator.from_input_examples(eval_triplets(test_functions))
+    dev_evaluator = OldFitScalarEvaluator(TripletEvaluator.from_input_examples(eval_triplets(test_functions)))
```
fleurvanl, line 137: This might be worth a comment. Do I understand correctly that we tripletify test_functions using eval_triplets; then push that into TripletEvaluator.from_input_examples in order to... reformat? and then push that result into OldFitScalarEvaluator to pick the correct metric? But where do the metrics get calculated?

fleurvanl (approved):

asmtransformers/scripts/finetune.py
ranieri: yes :P

The evaluator is called by the Trainer during old_fit.

PR #93, opened 2026-07-14
use config from BERT large

AstroVasseL:
Use config from BERT-large instead of small bert, based on https://huggingface.co/google-bert/bert-large-uncased/blob/main/config.json


asmtransformers/asmtransformers/models/multilingual_asmbert/config_large.json
```diff
@@ -0,0 +1,24 @@
+{
```
fleurvanl, line 2: might be nice to add a comment to say you need to specifiy the --config parameter in pretrain.py in order to use this config file and not the general config.json?

asmtransformers/asmtransformers/models/multilingual_asmbert/config.json
fleurvanl, line 23: is this specific for this PR or was this an error that you corrected on the side? Because I assume our vocab size is stable regardless of whether we train a regular or a large bert model?

fleurvanl (approved):

asmtransformers/asmtransformers/models/multilingual_asmbert/config.json
ranieri: Yes, this was a separate error, I'm guessing it was mine. I fixed manually for the pretrain runs and never committed it.

PR #94, opened 2026-07-16
inference takes embeddings from token ids instead of tokenizing the data

fleurvanl:
on testset of 400.000 functions, inference with token ids to embeddings took 12 minutes while inference with cfg to embeddings took 13 minutes. Small difference, but as we have about 20 million functions it will add up in the end

### DIFF
--- a/asmtransformers/asmtransformers/models/embedder.py
+++ b/asmtransformers/asmtransformers/models/embedder.py
@@ -46,7 +46,7 @@ class ASMEmbedder:
 
         embeddings = []
         with torch.no_grad():
-            for batch in batched(sentences, batch_size, strict=False):
+            for batch in tqdm(batched(sentences, batch_size, strict=False)):
                 inputs = self.tokenizer(batch, architecture=architecture)
                 inputs = {key: value.to(self.device) for key, value in inputs.items()}
                 outputs = self.model(**inputs)
--- a/asmtransformers/asmtransformers/models/embedder.py
+++ b/asmtransformers/asmtransformers/models/embedder.py
@@ -60,3 +60,39 @@ class ASMEmbedder:
         if convert_to_numpy:
             return embeddings.numpy().astype(np.float32, copy=False)
         return embeddings
+
+    def turn_into_tensors(self, batch, device):
+        # there might be a smarter way to do this, suggestions welcome!
+        features = {
+            'input_ids': torch.tensor([example[0] for example in batch], dtype=torch.long).to(self.device),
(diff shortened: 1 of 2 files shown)


asmtransformers/asmtransformers/models/embedder.py
```diff
@@ -60,3 +61,39 @@ def encode(
         if convert_to_numpy:
             return embeddings.numpy().astype(np.float32, copy=False)
         return embeddings
+
+    def turn_into_tensors(self, batch, device):
+        # there might be a smarter way to do this, suggestions welcome!
+        features = {
+            'input_ids': torch.tensor([example[0] for example in batch], dtype=torch.long).to(self.device),
+            'attention_mask': torch.tensor([example[1] for example in batch], dtype=torch.long).to(self.device),
```
akaIDIOT: Parameter `device` isn't used, and we've got access to `self.device`, so why not ditch the parameter (and omit it at the call site :)).


```suggestion
    def turn_into_tensors(self, batch):
        # there might be a smarter way to do this, suggestions welcome!
        features = {
            'input_ids': torch.tensor([example[0] for example in batch], dtype=torch.long).to(self.device),
            'attention_mask': torch.tensor([example[1] for example in batch], dtype=torch.long).to(self.device),
```

asmtransformers/scripts/inference.py
```diff
@@ -11,7 +11,9 @@ def main(data_folder, output_folder, model_path):
     print('Load model')
     model = ASMEmbedder.from_pretrained(model_path)
     print('Start creating embeddings')
-    embedded_functions = Dataset.from_dict({'embeddings': model.encode(eval_functions['cfg'])})
+    embedded_functions = Dataset.from_dict(
+        {'embeddings': model.get_embeddings(eval_functions['input_ids'], eval_functions['attention_mask'])}
+    )
```
akaIDIOT: Inference now assumes these columns are available. Alternatively, `inference.py` could check what columns are available and support both? Would that make sense or be silly?

asmtransformers/asmtransformers/models/embedder.py
```diff
@@ -60,3 +61,39 @@ def encode(
...
+        }
+        return features
+
+    def get_embeddings(
```
akaIDIOT: As discussed IRL: this duplicates a good chunk of code from `encode()`. Let's see what we can come up with a way to marry the two :)

akaIDIOT: On second thought, this can be fixed later, let's not go to Camelot, 'tis a silly place 🤡

akaIDIOT (approved):

PR #95, opened 2026-07-21
Add user_id as an optional field to database.add_function()

akaIDIOT:
Pending discussion with others.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/citatio/tests/test_db.py
+++ b/citatio/tests/test_db.py
@@ -3,6 +3,8 @@ from itertools import pairwise
 import numpy
 import pytest
 
+from citatio.db import PostgreSQLDatabase, SQLiteDatabase
+
 
 @pytest.fixture
 async def filled_database(database, functions, embeddings):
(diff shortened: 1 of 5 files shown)


citatio/tests/test_db.py
```diff
@@ -26,22 +26,37 @@ async def test_add_duplicate(database, functions, embeddings):
...
-            function['binary_name'],
-            function['binary_sha256'],
+            binary_name=function['binary_name'],
+            binary_sha256=function['binary_sha256'],
         )
         == function_id
     )
 
 
+async def test_add_binary_optional(database, functions, embeddings):
```
AstroVasseL:
```suggestion
async def test_add_binary_optional_userid(database, functions, embeddings):
```

citatio/tests/test_db.py
```diff
@@ -26,22 +26,37 @@ async def test_add_duplicate(database, functions, embeddings):
...
+    function = functions[-1]
+    embedding = embeddings[function['name']]
+
+    await database.add_function(function['name'], function['cfg'], embedding, user_id='nobody@asmembedder.local')
```
AstroVasseL: Good that this is tested. But the database never 'returns' user_id. Would be a nice-to-have to have a test that confirmes this being saved in postgres instance.

AstroVasseL (approved):

citatio/tests/test_db.py
akaIDIOT: The test name refers to the binary_name and binary_sha256 being optional. Let's add a separate test for the user_id.

citatio/tests/test_db.py
akaIDIOT: As with the other comment, let's add a separate test for this and remove the user_id here.

PR #98, opened 2026-07-21
Feature/eval ready

fleurvanl:


asmtransformers/scripts/evaluation.py
```diff
@@ -3,6 +3,7 @@
 import datetime as dt
 import os.path
```
AstroVasseL: I know it is not part of this pr, but I personally do not see a reason not to use `from pathlib import Path` here.

asmtransformers/scripts/evaluation.py
```diff
@@ -266,15 +264,46 @@ def get_parser():
...
+                args.output_path,
+                args.pool_size,
+                args.static_pool,
+                architecture,
+                args.seed,
+                args.repeats,
+            )
+            # right now we only report the MRRs and not the accuracies as they are usually zero
+            mrrs = np.array([row[2] for row in aggregate_rows])
+            # if there are multiple repeats, there will be multiple mrrs so we take the mean of those
```
AstroVasseL, line 286: Wait, I thought we decided on not having multilpe repeats but one static pool with the same seed? Does this really need to be in here? I foresee it only causing confusion later.

asmtransformers/scripts/evaluation.py
```diff
@@ -242,19 +245,14 @@ def run_tests(data_folder, output_path, pool_size, static_pool, architecture, se
...
+    return aggregate_rows
 
 
 def get_parser():
     parser = argparse.ArgumentParser(description='evaluation')
     parser.add_argument('input_path', type=str, help='the path to the test data')
     parser.add_argument('output_path', type=str, help='the path to write the final scores to')
     parser.add_argument('--pool-size', type=int, help='the poolsize to pick the positive example from')
-    parser.add_argument('--architecture', type=str, help='only use examples from specified architecture')
     parser.add_argument('--seed', type=int, help='seed random evaluation sampling')
```
AstroVasseL: Would it be wise to give a default seed here, since we will be comparing our models on one seed?

asmtransformers/scripts/evaluation.py
```diff
@@ -266,15 +264,46 @@ def get_parser():
...
+            aggregate_rows = run_tests(
+                args.input_path,
+                args.output_path,
+                args.pool_size,
+                args.static_pool,
+                architecture,
+                args.seed,
+                args.repeats,
+            )
+            # right now we only report the MRRs and not the accuracies as they are usually zero
```
AstroVasseL, line 283: If the accuracy is not reported then why would we calculate it in run_tests? 
I would argue that we should report it here. If it does is always zero it should probably be replaced with another metric, but the code should report it. If one model all of a sudden does have a nonzero accuracy, I think it should be reported also.

asmtransformers/scripts/evaluation.py
```diff
@@ -209,14 +209,17 @@ def run_tests(data_folder, output_path, pool_size, static_pool, architecture, se
...
+        architecture = None
 
     print('\ngenerate test_pools\n')
     test_functions = load_test_functions(data_folder, architecture)
     anchor_rng = random.Random(seed)
     anchor_pairs = generate_anchor_pos_pairs(test_functions, anchor_rng)
 
     model_name = data_folder.split('/')[-1]
-    output_file = f'{model_name}-{pool_size}-{static_pool}-{timestamp()}'
+    output_file = f'{timestamp()}-{model_name}-{architecture}-{pool_size}-{static_pool}'
```
AstroVasseL, line 222: Is static pool a boolean value? And in our current implementation, do we always have a static pool?
If yes, do we want it in the output_file name?

AstroVasseL (approved):

asmtransformers/scripts/evaluation.py
```diff
@@ -209,14 +209,17 @@ def run_tests(data_folder, output_path, pool_size, static_pool, architecture, se
         raise ValueError('repeats must be at least 1')
     if repeats > 1 and not static_pool:
         raise ValueError('repeats greater than 1 are only supported with --static-pool')
+    # architecture is a filter; if we want to evaluate all architectures at once, we do not filer the dataset
```
akaIDIOT, line 212: Good comment 👌🏻

asmtransformers/scripts/evaluation.py
```diff
@@ -266,15 +264,46 @@ def get_parser():
...
+                results_per_architecture['riscv64'],
+                results_per_architecture['i386'],
+                results_per_architecture['all'],
+            )
+        )
+
+    with open(os.path.join(args.output_path, output_file + '-parameters.txt'), 'w') as file:
+        file.write(
+            f'{args.input_path=},\n {args.output_path=},\n {args.pool_size=},\n {args.static_pool=},\n'
+            f' {args.seed=},\n {args.repeats=},\n'
```
akaIDIOT: Interesting format. While likely just informative, why the space before all but the first value?

asmtransformers/scripts/evaluation.py
```diff
@@ -266,15 +264,46 @@ def get_parser():
...
+                results_per_architecture['arm64'],
+                results_per_architecture['amd64'],
+                results_per_architecture['riscv64'],
+                results_per_architecture['i386'],
+                results_per_architecture['all'],
+            )
+        )
+
+    with open(os.path.join(args.output_path, output_file + '-parameters.txt'), 'w') as file:
+        file.write(
```
akaIDIOT: Using `pathlib`, this can be simpler, using `Path.write_text(...)`

```python
(args.output_path / f'{output_file}-parameters.txt').write_text(f'{args.input_path=}...')
```

(not that this assumes `args.output_path` is a `pathlib.Path`, which is easy to do with `type=Path` instead of `type=str`)

asmtransformers/scripts/evaluation.py
fleurvanl: after discussion we decided to leave it like this

asmtransformers/scripts/evaluation.py
fleurvanl: decided to keep repeats after discussion

Issue #91, opened 2026-07-03
Add a way to deal with schema changes


akaIDIOT:
Current implementation of either database API will run the schema script on startup, creating tables unless they already exist. If we change the schema, this will break for existing data. Some schema upgrader / migration thing would be nice.

PR #99, opened 2026-07-22
Enforce unique label per function per user, unless user is anonymous

akaIDIOT:


citatio/citatio/db.py
```diff
@@ -148,6 +156,8 @@ async def add_function(self, name, cfg, embedding, *, user_id=None, binary_name=
                     labels (function_id, label, user_id, binary_name, binary_sha256)
                 VALUES
                     ($1, $2, $3, $4, $5)
+                ON CONFLICT (function_id, user_id) DO UPDATE
+                    SET label = $2, binary_name = $4, binary_sha256 = $5
```
akaIDIOT, line 160: Could we use named parameters here too? The positional ones feel clunky.

citatio/citatio/schema-sqlite.sql
```diff
@@ -30,5 +30,9 @@ CREATE TABLE IF NOT EXISTS labels (
     binary_name TEXT DEFAULT NULL,
     binary_sha256 TEXT DEFAULT NULL,
 
-    FOREIGN KEY (function_id) REFERENCES functions(id)
+    FOREIGN KEY (function_id) REFERENCES functions(id),
+    -- enforce unique function labels per user
```
AstroVasseL, line 34: Out of interest, do these changes in our sqlite schema make our code not backwards compatible? Would we need a database migration of the 'old' sqlite database to be compatible with our latest version?

citatio/citatio/db.py
AstroVasseL: do you mean like 
```suggestion
                    (:function_id, :label, :user_id, :binary_name, :binary_sha256)
                ON CONFLICT (function_id, user_id) DO UPDATE
                    SET label = :label, binary_name = :binary_name, binary_sha256 = :binary_sha256
```

citatio/tests/test_db.py
```diff
@@ -59,6 +59,44 @@ async def test_add_binary_fields_optional(database, functions, embeddings):
...
+    results = await database.search_function(embedding)
+    assert len(results) == 2
+    assert {result['function'] for result in results} == {function['name'], 'new_name'}
+    assert all(result['similarity'] == pytest.approx(1.0) for result in results)
+
+
+async def test_add_label_user_overwrite(database, functions, embeddings):
+    function = functions[-2]
+    embedding = embeddings[function['name']]
+    await database.add_function(function['name'], function['cfg'], embedding, user_id='henk@reverse-engineering.nl')
```
AstroVasseL, line 91: nice

AstroVasseL (approved):

citatio/citatio/schema-sqlite.sql
akaIDIOT: Correct, an existing database would not work with this code. This issues has been deferred to #91 O:)

PR #101, opened 2026-07-22
Add user identity as an IDToken

akaIDIOT:


citatio/citatio/api.py
```diff
@@ -42,15 +67,17 @@ async def add_function(
     name: Annotated[str, Body()],
     cfg: Annotated[ControlFlowGraph, Body()],
     architecture: str = 'arm64',
-    binary_name: Annotated[str, Body()] = None,
-    binary_sha256: Annotated[str, Body()] = None,
+    binary_name: Annotated[str | None, Body()] = None,
+    binary_sha256: Annotated[str | None, Body()] = None,
+    id_token: IDToken | None = Depends(authenticated_user),  # noqa: B008 (not a function call, just the FastAPI way™)
```
akaIDIOT: This should be done in the FastAPI way™, comment is wrong O:)


```suggestion
    id_token: Annotated[IDToken | None, Depends(authenticated_user)],
```

citatio/citatio/db.py
```diff
@@ -17,11 +21,9 @@ async def add_function(
         user_id: str | None = None,
         binary_name: str | None = None,
         binary_sha256: bytes | None = None,
-    ):
-        pass
+    ): ...
```
akaIDIOT, line 24: TIL: using `...` instead of `pass` in a stub functions saves a line per function 👏🏻

citatio/citatio/api.py
```diff
@@ -42,15 +67,17 @@ async def add_function(
...
+    id_token: Annotated[IDToken | None, Depends(authenticated_user)] = None,
 ):
     embedding = request.app.state.model.encode(str(cfg), architecture=architecture)
     await request.app.state.database.add_function(
         name,
         cfg,
         embedding,
-        user_id=None,  # TODO: use authenticated user id when available
+        # user_id is optional, use subject identifier from token if available
+        user_id=id_token.sub if id_token else None,
```
judithvdw, line 81: Not entirely sure how to formulate this, but Is there a way to get some sort of proxy for a user or a session if there is no id_token? Or is it just fine that all the anonymous ones are exactly the same?

citatio/citatio/api.py
```diff
@@ -3,36 +3,61 @@
...
         yield
 
 
+async def authenticated_user(request: Request) -> IDToken | None:
+    if auth := request.headers.get('Authorization'):
+        # authorization header available, let auth create a token from it
+        return request.app.state.authenticate_user(auth)
+    else:
+        # anonymous request, no token
+        return None
```
judithvdw, line 59: I'm  not sure if I fully understand it. Authentication (OIDC) is mandatory, but you are allowed to be anonymous. Is that correct?

judithvdw (approved):
As far as I can tell it looks good, have a few questions, might be useful to have a bit more explanation about those somewhere.

citatio/citatio/api.py
akaIDIOT: This particular thing is due to wanting to allow anonymous access, which would be signaled through not sending a token (leaving the header empty). But the `authenticate_user` function, when called, will immediately raise an HTTP 401 because there is no token in the header. By checking the header first, and allowing the *absence* of that header, we're able to

- allow anonymous access by leaving the header empty;
- allow authenticated access, delegated to a 3rd party lib that will validate the token in the authorization header.

I'll add some additional docs / comments on the how and why, though #104 expands on this.

citatio/citatio/api.py
akaIDIOT: The current setup is stateless / has no session. The only proxy we have (when there's no IDToken) is an IP address, not sure that's very useful 🤔

akaIDIOT: Added a task to include a timestamp with a label, giving someone checking the database a chance to find what they were looking for.

citatio/citatio/api.py
akaIDIOT: Expanded the line comments a bit, though #104 will likely need some along those lines too.

PR #104, opened 2026-07-23
Support unvalidated explicit user as identity

akaIDIOT:


citatio/citatio/api.py
```diff
@@ -37,12 +45,9 @@ async def lifespan(app: FastAPI):
...
 
-    if config.auth:
-        # create an OIDC Authorization header → IDToken function from the configured authentication settings
-        app.state.authenticate_user = get_auth(**config.auth)
-    else:
-        # config.auth either not set or explicitly turned off, raise exception on presence of Authorization header
-        app.state.authenticate_user = _auth_unavailable
+    app.state.authenticate_user = resolve_auth(**config.auth)
+    # TODO: determine from config, apply stricter defaults
+    app.state.identification_modes = {'anonymous', 'supplied', 'oidc'}
```
judithvdw: The 'supplied' comes a bit out of nowhere in this MR, I can't find what is does or when I would want to use this. Either a comment in the code, or in the readme would be useful.

judithvdw (approved):

citatio/citatio/api.py
akaIDIOT: As discussed IRL, made it `client_supplied` for now, documented its particulars (and meaning) in `citatio/README.md`.

PR #105, opened 2026-07-24
Tests/evaluation

fleurvanl:


asmtransformers/scripts/evaluation.py
```diff
@@ -174,7 +179,7 @@ def calculate_one_rank(row):
     # If we fall through the for loop without hitting the return statement,
     # we are worse than all the negatives. Our rank is therefore
     # equal to the number of negatives.
-    return len(similarities)
+    return len(similarities) + 1
```
akaIDIOT: Code is hard to understand with the comment above it, equal to the number of negatives, but:

- we're counting similarities;
- we're returning a count + 1

asmtransformers/tests/test_evaluation.py
```diff
@@ -0,0 +1,89 @@
...
+        'anchor': {'embeddings': [0, 1, 2]},
+        'pos': {'embeddings': [7, 8, 9]},
+        'negs': np.array([[4, 5, 6], [1, 2, 3]]),
+    }
+
+
+def test_calculate_one_rank(rank1, rank2, rank3):
+    assert calculate_one_rank(rank1) == 1
+    assert calculate_one_rank(rank2) == 2
+    assert calculate_one_rank(rank3) == 3
```
akaIDIOT, line 44: Just from reading the code, it's hard to see what makes these rank 1, 2 and 3. Would a line comment at the fixtures help with understanding what makes it a rank 1, 2 or 3?

asmtransformers/tests/test_evaluation.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def test_calculate_all(rank1, rank2, tmp_path):
```
akaIDIOT, line 47: By using the same file name thrice, this test inadvertently tests that these files get overwritten. Would it make sense to also verify the content of the file after it gets written?

asmtransformers/tests/test_evaluation.py
```diff
@@ -0,0 +1,89 @@
...
+
+@pytest.fixture
+def rng():
+    return random.Random(10)
+
+
+def test_generate_anchor_pos_pairs(dataset, rng):
+    anchors, positives, anchor_labels, anchor_cfgs, pos_cfgs = generate_anchor_pos_pairs(dataset, rng, num_pairs=3)
+    print(anchors)
+    print(positives)
```
akaIDIOT: Think these should be removed?

asmtransformers/tests/test_evaluation.py
```diff
@@ -0,0 +1,89 @@
...
+            'label': ['return', 'return', 'jump', 'jump', 'move', 'move', 'add', 'add'],
+            'cfg': [
+                'ret',
+                'ret',
+                'j 0x32',
+                'c.j 0x02',
+                'c.mv a3,a5',
+                'c.mv s8,a5',
+                'c.addi4spn s0,sp,0x30',
+                'addi a5,s0,-0xb0',
```
akaIDIOT: I'm assuming this test dataset will map 8 function names to 8 cfgs? Altough the format of a cfg in a dataset is `str`, I think it could help if these cfgs look like cfgs, iirc a list of *[offset, [instruction, ...]]*. Would it make sense to format these along those lines?


```python
                '[[0, [ret]]]',
                '[[1, [ret]]]',
                ...
```

asmtransformers/tests/test_evaluation.py
```diff
@@ -0,0 +1,89 @@
...
+            ],
+        }
+    )
+
+
+@pytest.fixture
+def rng():
+    return random.Random(10)
```
akaIDIOT, line 82: Lovely to see something other than 42 😄

asmtransformers/scripts/evaluation.py
```diff
@@ -106,6 +106,11 @@ def generate_anchor_pos_pairs(dataset, rng):
             # same content, reject
             rejected += 1
             continue
+        if anchor['cfg'] in anchor_cfgs:
+            # this cfg has already been used as anchor
+            # it can still be used as positive for another anchor with the same label
+            rejected += 1
```
akaIDIOT, line 120: Noticed before that `rejected` is never used in this code. Maybe not something to change / fix here, but curious nonetheless 🤔

akaIDIOT (approved):

asmtransformers/scripts/evaluation.py
fleurvanl: Do you think we should print something at the end about how many pairs were rejected?

akaIDIOT: Or just not count them at all, what's the use in counting if we're not going to do anything with that value 🤷🏻

asmtransformers/scripts/evaluation.py
fleurvanl: Better like this?

asmtransformers/scripts/evaluation.py
akaIDIOT: Wait, it gets logged somewhere, it seems... Maybe it's not that useless after all...

asmtransformers/scripts/evaluation.py
akaIDIOT: More better, but ... why does the comment mention "number of negatives", but the line below it does `len(similarities)`? Would it not make sense to do `len(negs)`? If that's incorrect, the comment is still not helping 😅

PR #106, opened 2026-07-24
Enable loading fake embedder, avoids retrieving model from HF hub for tests

akaIDIOT:


citatio/tests/test_api.py
```diff
@@ -60,4 +55,4 @@ def test_search_unknown(client, functions):
     assert len(results) == 3
     for result in results:
         # nothing matches exactly, nothing should come back < 0.0
-        assert 0.1 < result['similarity'] < 0.9
+        assert 0.0 < result['similarity'] < 1.0
```
judithvdw, line 61: I'm assuming this is because the fake embedder returns a different range of results than the actual model on HF?

citatio/citatio/models.py
```diff
@@ -38,3 +40,15 @@ def __iter__(self):
     def __str__(self):
         # __str__ is invoked on str-coercion, str(cfg) would result in a JSON-encoded list of blocks
         return json.dumps(self.blocks)
+
+
+class FakeEmbedder:
```
judithvdw: Would TestEmbedder be better? I'm not sure if this one is 'fake' it is just... not useful in practice.

judithvdw (approved):

citatio/tests/test_api.py
akaIDIOT: I don't know why it was [0.1, 0.9], considering the comment above it mentions 0.0 🤷🏻 

As the fake / test embedder makes nonsensical vectors, I can't guarantee the distance being within certain boundaries, other than the range of the value, which would be [0.0, 1.0].

citatio/citatio/models.py
akaIDIOT: That's a better name 👍🏻 

In the meantime, looking at ways to move this implementation into test code, if we can express the embedder as a `Depends`, without loading it for every request. Online clanker says the lifecycle management we're currently using is the way to implement this, but let's see if we can turn this into a (threadsafe) lazy loader so we can override it for tests...

akaIDIOT: Created a new task to move the model loader into a Depends, leaving the (renamed) `TestEmbedder` in production code for now, accompanied by a TODO :)

PR #107, opened 2026-07-24
Feature/save eval pairs

fleurvanl:
This probably should have been two PRs:
1. testing for evaluation.py, which is almost all code in evaluation.py and test_evaluation.py
2. creating separate scripts in order to:
        - save the anchor-pos-negative pairs WITHOUT their embeddings
        - embed this previously saved dataset using a given model
        - evaluate that given model
    this all happens in eval-dataset.py and evaluate-static-dataset.py

### DIFF
--- a/asmtransformers/scripts/eval-datasets.py
+++ b/asmtransformers/scripts/eval-datasets.py
@@ -18,6 +18,7 @@ def get_parser() -> argparse.ArgumentParser:
 
 def main(data_folder, output_folder, pool_size, seed):
     anchor_rng = random.Random(seed)
+    neg_rng = random.Random(seed + 1)
     architectures = ['arm64', 'amd64', 'riscv64', 'i386', 'all']
     for architecture in architectures:
         print(f'creating eval data for {architecture}')
@@ -25,9 +26,7 @@ def main(data_folder, output_folder, pool_size, seed):
         anchors, positives, anchor_labels, anchor_cfgs, pos_cfgs = generate_anchor_pos_pairs(
             test_functions, anchor_rng, num_pairs=1000
         )
-        neg_pools = generate_neg_pool(
-            pool_size, test_functions, anchor_labels, anchor_cfgs, pos_cfgs, random.Random(seed + 1)
-        )
+        neg_pools = generate_neg_pool(pool_size, test_functions, anchor_labels, anchor_cfgs, pos_cfgs, neg_rng)
         Path.mkdir(Path(output_folder, architecture), parents=True, exist_ok=True)
         with Path(output_folder, architecture, 'eval_data.pkl').open('wb') as f:
             # we save only cfgs for anchors and positives; from neg_pools we are already holding only cfgs
--- a/asmtransformers/scripts/eval-datasets.py
+++ b/asmtransformers/scripts/eval-datasets.py
@@ -1,17 +1,16 @@
 import argparse
-import os
 import pickle
 import random
 from pathlib import Path
 
-from scripts.evaluation import generate_anchor_pos_pairs, generate_triplets, load_test_functions, generate_neg_pool
+from scripts.evaluation import generate_anchor_pos_pairs, generate_neg_pool, load_test_functions
 
 
 def get_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(add_help=False)
     parser.add_argument('data_folder', type=str, help='folder with input data')
     parser.add_argument('output_folder', type=str, help='folder to leave output data')
-    parser.add_argument('--pool-size', type=int, help='the poolsize to pick the positive example from')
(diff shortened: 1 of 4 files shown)


asmtransformers/scripts/eval-datasets.py
```diff
@@ -0,0 +1,40 @@
...
+    anchor_rng = random.Random(seed)
+    architectures = ['arm64', 'amd64', 'riscv64', 'i386', 'all']
+    for architecture in architectures:
+        print(f'creating eval data for {architecture}')
+        test_functions = load_test_functions(data_folder, architecture)
+        anchors, positives, anchor_labels, anchor_cfgs, pos_cfgs = generate_anchor_pos_pairs(
+            test_functions, anchor_rng, num_pairs=1000
+        )
+        neg_pools = generate_neg_pool(
+            pool_size, test_functions, anchor_labels, anchor_cfgs, pos_cfgs, random.Random(seed + 1)
```
AstroVasseL: At first I was caught a bit off-guard by `random.Random(seed + 1)` here, but looking into the function i saw that it is the neg_pool rng. Would it be an idea to have, at the beginning of the function, an `anchor_rng = random.Random(seed)` and a `neg_rng = random.Random(seed+1)` to make it a bit more readable?

asmtransformers/scripts/evaluation-static-dataset.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def load_triplets(data_path):
+    with Path(data_path).open('rb') as file:
+        anchors, positives, neg_pools = pickle.load(file)
+    anchors = model.encode(anchors)
+    positives = model.encode(positives)
+    neg_pools = model.encode(neg_pools)
+
+    for i in range(len(anchors)):
```
AstroVasseL: why not an `for idx, (anchor,positive) in enumerate(zip(anchors, positives))` ?

AstroVasseL: Same effect but feels more pydantic :)

asmtransformers/scripts/evaluation-static-dataset.py
```diff
@@ -0,0 +1,97 @@
...
+                'mrr_arm64',
+                'acc_arm64',
+                'mrr_amd64',
+                'acc_amd64',
+                'mrr_riscv64',
+                'acc_riscv64',
+                'mrr_i386',
+                'acc_i386',
+                'mrr_all',
+                'acc_all',
```
AstroVasseL, line 78: Arent all these values `results_per_achitecture.keys()`? Could this also be `keys = results_per_architecture.keys();
writer.writerow(keys)`
and
`metrics = [results_per_architecture[k] for k in keys];
writer.writerow(metrics)` 

If this is not possible I would still prefer ['mrr'] and ['acc'] over [0] and [1] if ever the metrics change.

asmtransformers/scripts/evaluation.py
```diff
@@ -56,15 +56,15 @@ def generate_neg_pool(pool_size, dataset, anchor_labels, anchor_cfgs, pos_cfgs,
...
-        if len(neg_embeddings) == pool_size:
+        negs.append(neg)
+        if len(negs) == pool_size:
             break
-    if len(neg_embeddings) < pool_size:
-        raise ValueError(f'only {len(neg_embeddings)} eligible negative examples available for pool_size={pool_size}')
-    return np.array(neg_embeddings)
+    if len(negs) < pool_size:
+        raise ValueError(f'only {len(negs)} eligible negative examples available for pool_size={pool_size}')
+    return negs  # np.array(negs)
```
AstroVasseL: Should this line commend still be here?

asmtransformers/scripts/evaluation.py
```diff
@@ -141,7 +155,8 @@ def generate_triplets(dataset, anchor_pairs, pool_size, static_pool, rng):
 
 def load_test_functions(data_folder, architecture=None):
     dataset = datasets.load_from_disk(data_folder)  # .select(range(11000, 45000))
-    if architecture:
+    # if we want to evaluate all architectures, we do not need to filter the data
```
AstroVasseL, line 158: nice

asmtransformers/scripts/evaluation.py
```diff
@@ -157,7 +172,11 @@ def calculate_one_rank(row):
     :param row: anchor, pos, and list of negs
     :return The rank of the positive example (pos) in the pool.
     """
-    anchor, pos, negs = row['anchor']['embeddings'], row['pos']['embeddings'], row['negs']
+    anchor, pos, negs = (
+        row['anchor']['embeddings'],
+        row['pos']['embeddings'],
+        np.array([i['embeddings'] for i in row['negs']]),
```
AstroVasseL: Generally when I see i in a for loop i think it is index. Is that the case here? If not, can `i` get another name, maybe `neg`?

AstroVasseL (approved):

PR #110, opened 2026-07-28
Make REST paths more REST-like, use "label" as the name / identifier / pointer to a function everywhere

akaIDIOT:


citatio/citatio/api.py
```diff
@@ -122,12 +122,12 @@ async def add_function(
     )
 
 
-@app.post('/api/v1/search')
+@app.post('/api/v1/functions/search')
```
akaIDIOT, line 125: The world has multiple views on how to put search functions in REST (as in: it might be more logical if this was a GET), but one of the limiting factors in that is that the cfg would then end up in the URL's query part, which has a limit in length that we might be hitting. POST does not have this limitation, so let's keep that detail. At least the path now indicate the resource type that the search applies to (functions).

AstroVasseL: Agreed. Lets just say we do not GET the search values but actually POSTing our search query ;) .

AstroVasseL (approved):

PR #111, opened 2026-07-28
Expose auth configuration from REST

akaIDIOT:
At least the configured base url is needed by the client to go figure out where to get a token from. The client id might aid in finding the right configuration at `base_url/.well-known/oidc-configuration/`.

### DIFF
--- a/citatio/tests/test_api.py
+++ b/citatio/tests/test_api.py
@@ -70,6 +70,8 @@ def test_add_function_anonymous_not_allowed(monkeypatch, client, functions):
     monkeypatch.setattr(app.state, 'identification_modes', {'client_supplied'})
     response = client.post('/api/v1/functions', json=functions[0])
     assert response.status_code == 401
+    response = client.post('/api/v1/functions', json={**functions[1], 'user_id': 'MiniDane'})
+    assert response.status_code == 200
 
 
 def test_search_known(client, functions):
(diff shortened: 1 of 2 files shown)


citatio/tests/test_api.py
```diff
@@ -16,6 +16,38 @@ async def client(monkeypatch, database_env):
         yield client
 
 
+def test_get_auth_config(client):
+    response = client.get('/api/v1/auth')
+    assert response.status_code == 200
+    assert response.json() == {'anonymous': True, 'client_supplied': True, 'oidc': False}
```
AstroVasseL, line 22: From the README: 
"The authentication modes to support, any of anonymous, client_supplied and oidc;"

Is there also a world where there is a client_suppliet, non-anonymous authentication? And can that be tested?

AstroVasseL (approved):

citatio/tests/test_api.py
akaIDIOT: The three options are independent: `client_supplied` is by definition not anonymous, it's assuming there's a `user_id` inside the request that is supposed to just be trusted. 

Added a post + assert to `test_add_function_anonymous_not_allowed`, where only `client_supplied` is enabled, that shows an anonymous request is not accepted, while a request that includes a user id *is*.

akaIDIOT: The way I see it btw, is that a realish production env will only contain / support oidc here, the rest might be useful for testing or local (or even localhost) deployments.

PR #115, opened 2026-07-29
SLURM_script that runs pretrain.py

AstroVasseL:


asmtransformers/scripts/launch-pretrain.sh
```diff
@@ -0,0 +1,28 @@
...
+    --nnodes "$NNODES"
+    --nproc-per-node "$GPUS_PER_NODE" 
+    --rdzv-backend "c10d" 
+    --rdzv-endpoint "$MASTER_ADDR":"$MASTER_PORT" 
+)
+
+TRAINIG_ARGS=(
+    --data "/data/dataset"
+    --tokenizer "/data/tokenizer"
+    --config "/app/asmtransformers/models/multilingual_asmbert/config_large.json"
```
akaIDIOT: I'm assuming that these paths correspond to some defaults that are defined elsewhere? Can we include such a pointer here as a comment or something?

akaIDIOT (commented):
Not familiar with the SLURM / torchrun setup this targets, would it help to rope in someone who's more in touch with that? :)

asmtransformers/scripts/launch-pretrain.sh
AstroVasseL: Yes! When creating the container, the data, tokenizer and config need bind on these locations.

asmtransformers/launch-pretrain.sh
```diff
@@ -0,0 +1,28 @@
...
+    --rdzv-endpoint "$MASTER_ADDR":"$MASTER_PORT" 
+)
+
+TRAINIG_ARGS=(
+    --data "/data/dataset"
+    --tokenizer "/data/tokenizer"
+    --config "/app/asmtransformers/models/multilingual_asmbert/config_large.json"
+    --epoch 19
+    --batch-size 192
+    --gradient-accumulation-steps 1
```
judithvdw: If we trained with a larger batch size on Gracie should we not train with some gradient accumulation here? To emulate the same effective batch size?

asmtransformers/scripts/launch-pretrain.sh
```diff
@@ -0,0 +1,28 @@
...
+    --epoch 19
+    --batch-size 192
+    --gradient-accumulation-steps 1
+    --mlm-prob 0.15
+    --save-steps 5000
+    --dataloader-num-workers 4
+    --bf16
+    --tf32
```
judithvdw, line 22: with the lines above, some of these are now the same as the default in peretrain.py but explicitly mentioned, where as some arguments are just left on default without mentioning. 

I'm assuming because it is the ones we actually want to experiment with, but might be good to have a comment on that in the script here.

judithvdw: Also, maybe good to in this pull request also add some notes about this shell script in the readme, what it is for and how to work with it (like having to edit the arguments here).

Most stuff related to that can be in the separate readme on surf/slurm, but some mention might be good.

asmtransformers/pretrain.Containerfile
```diff
@@ -1,6 +1,6 @@
 FROM python:3.13
 ARG PIP_PYPI_URL=https://pypi.org/simple
-COPY --parents asmtransformers scripts pyproject.toml pdm.lock README.md /app
+COPY --parents asmtransformers scripts launch-pretrain.sh pyproject.toml pdm.lock README.md /app
```
judithvdw: I know I am the one who originally put launch-pretrain.sh in the root in stead of in scripts (because I was trying stuff out and wanted to make sure it worked from the root without thinking about working directories). But not actually sure if it is necessary to have it there. @akaIDIOT usually has opinions about these things, I don't mind ;)

judithvdw (approved):
Looks good. Do you think, as we need to have this here anyways, we might as well put the other job script in the repository here too?

asmtransformers/scripts/launch-pretrain.sh
judithvdw: One other thing I was wondering, is there a way to pass these on from the script outside the container? Because now every time you want to do a different experiment (like MLM40) you have to make and push a new container. Whereas if the non-default arguments can be passed on from the job script, you would only have to make a new container. This probably still leaves you with some experiments you do have to build a new container for (Bertlarge vs normal maybe?), but for arguments parameterised in the cli of pretrain.py it feels like it should be possible?

AstroVasseL: This is difficult, since variables in bash always get passed on as strings. Therefore, anything that is a float of int will be a string until an arithmic operation will be performed on it. That is why I chose to put it here.

asmtransformers/pretrain.Containerfile
AstroVasseL: I don't mind putting it in the /scripts/ folder, I think one could argue it fits there.

akaIDIOT: It's quite literally a script, as there's no external tool that misbehaves when this is not in the root, I'd say put it in scripts? :)

asmtransformers/scripts/launch-pretrain.sh
akaIDIOT: And will we include a comment or something with that information here for future reference? :)

PR #116, opened 2026-07-29
Fix deprecated pretrain syntax

AstroVasseL:
This PR fixes the message: [transformers] warmup_ratio is deprecated and will be removed in v5.2. Use `warmup_steps` instead. message in transformers 
when running `pretrain.py`.


AstroVasseL:
Warmup_steps either accepts an int or a float between 0 and 1, so the expected behaivour will be the same: https://github.com/huggingface/transformers/pull/41326#issuecomment-3397773827

asmtransformers/scripts/pretrain.py
```diff
@@ -321,7 +321,7 @@ def build_arg_parser():
     parser.add_argument('--logging-steps', type=int, default=100, help='number of update steps between two logs')
     parser.add_argument('--mlm-prob', type=float, default=0.15, help='probability of a token/word to be masked')
     parser.add_argument('--learning-rate', type=float, default=1e-4, help='learning rate')
-    parser.add_argument('--warmup-ratio', type=float, default=0.06, help='warmup ratio for the learning-rate scheduler')
+    parser.add_argument('--warmup_steps', type=float, default=0.06, help='warmup steps for the learning-rate scheduler')
```
akaIDIOT: All other command line options (correctly, imo) use dashed separators, this should too, no?


```suggestion
    parser.add_argument('--warmup-steps', type=float, default=0.06, help='warmup steps for the learning-rate scheduler')
```

akaIDIOT (commented):
Just curious: what deprecated this, torch? Since when?

akaIDIOT (approved):

AstroVasseL:
> Just curious: what deprecated this, torch?
transformer.
> Since when?
It will be removed in transformers v5.2.

PR #120, opened 2026-08-04
Limit architecture to supported values in citatio

akaIDIOT:


AstroVasseL (approved):
I did not know StrEnum classes existed, cool

asmtransformers/asmtransformers/models/__init__.py
```diff
@@ -1,5 +1,13 @@
+from enum import StrEnum
 from importlib import resources
 
 
+class Architecture(StrEnum):
```
fleurvanl: is this specifically StrEnum because of the asyncpg issue? in that case could you put a comment about that?

citatio/citatio/api.py
```diff
@@ -126,7 +127,7 @@ async def add_function(
     request: Request,
     label: Annotated[str, Body()],
     cfg: Annotated[ControlFlowGraph, Body()],
-    architecture: Annotated[str, Body()] = 'arm64',
+    architecture: Annotated[Architecture, Body()],
```
fleurvanl, line 130: Does this mean we don't default to arm64 anymore? I believe we list that somewhere in the architecture.md  so in that case that needs changing

fleurvanl (approved):

asmtransformers/asmtransformers/models/__init__.py
akaIDIOT: It is not, but the fact that this isn't clear does deserve a comment. It's main purpose was to define a type that we could import and use in `citatio.api` to require architecture to be one of these four values, causing `fastapi` to create HTTP errors if someone sends a request with `"architecture": "mips"`, for example.

akaIDIOT: Added a docstring to `Architecture`, used it in a few additional places.

citatio/citatio/api.py
akaIDIOT: Ah, good point, removed the `='arm64'` default in lots of places, I believe all the tests still pass, but the argument is now required everywhere :)

citatio/citatio/db.py
```diff
@@ -88,7 +86,7 @@ async def add_function(
         binary_name=None,
         binary_sha256=None,
     ):
-        if architecture not in SUPPORTED_ARCHITECTURES:
+        if architecture not in Architecture:
             raise ValueError(f'unsupported architecture: {architecture}')
```
akaIDIOT: Hmm, could this be made simpler with an `Architecture(architecture)` or something along those lines? 🤔

akaIDIOT: It can!

PR #121, opened 2026-08-06
update citatio default model

AstroVasseL:


citatio/citatio/api.py
```diff
@@ -14,7 +14,7 @@
 
 
 SUPPORTED_AUTH_MODES = frozenset({'anonymous', 'client_supplied', 'oidc'})
-DEFAULT_MODEL = 'NetherlandsForensicInstitute/ARM64BERT-embedding'
+DEFAULT_MODEL = 'NetherlandsForensicInstitute/Multilingual-ASMBERT'
```
akaIDIOT: Don't currently have cuda locally, so not sure I can easily test this, let's try :)

citatio/citatio/api.py
akaIDIOT, line 1: The default also needs to be changed in the `defaults.toml` file. 

There should have been a way where this string is only the default in one place, sorry 🙈

AstroVasseL: Could we load DEFAULT_MODEL from defaults.toml?

citatio/citatio/api.py
AstroVasseL:
```suggestion
DEFAULT_MODEL = confidence.loads(resources.read_text('citatio', 'defaults.toml'), format=confidence.TOML).hf
```
Would this work? Would you think this is better?

akaIDIOT: Think it's running on the CPU, but trying silly things from the `/docs` API docs seems to work:

```plain
1it [00:00,  8.01it/s]
 ▕  127.0.0.1:38956 - "POST /api/v1/functions HTTP/1.1" 200
1it [00:00,  7.11it/s]
 ▕  127.0.0.1:39790 - "POST /api/v1/functions HTTP/1.1" 200
1it [00:00,  7.97it/s]
 ▕  127.0.0.1:50500 - "POST /api/v1/functions HTTP/1.1" 200
1it [00:00,  7.80it/s]
 ▕  127.0.0.1:56362 - "POST /api/v1/functions/search HTTP/1.1" 200
```

Not sure why there's a tqdm instead of something loggy, but at least it's embedding and the search is coming back with 'sane' results for my 'insane' tests 👍🏻

akaIDIOT: Hmm, to be honest, it might be better to remove `DEFAULT_MODEL` altogether, lemme check where we're using that constant...

akaIDIOT: We're not using it anywhere 🤡 How about we ditch the constant here and rely on `defaults.toml` as the single source of defaults 😆 


```suggestion
```

fleurvanl (approved):

PR #125, opened 2026-08-13
Experiment: Support intel xpu device for the model side of things

akaIDIOT:


asmtransformers/asmtransformers/models/__init__.py
```diff
@@ -1,5 +1,22 @@
...
+@cache
+def default_device():
+    import torch
+
+    if torch.cuda.is_available():
+        return torch.device('cuda')
+    if torch.xpu.is_available():
+        return torch.device('xpu')
+    if torch.mps.is_available():
+        return torch.device('mps')
```
akaIDIOT, line 18: This is wider than it really needs to be, but it *might* just work on macos now too.

asmtransformers/asmtransformers/models/embedder.py
```diff
@@ -14,16 +15,22 @@ def __init__(self, model, tokenizer, *, device=None, normalize_embeddings=True):
...
 
     @classmethod
     def from_pretrained(
         cls, model_name_or_path, *, model_args=None, tokenizer_args=None, device=None, normalize_embeddings=True
     ):
+        device = device or default_device()
+        model_args = model_args or {}
+        if device.type == 'xpu':
+            # SPDA might cause runtime errors on xpu, default to eager attention to avoid it
+            model_args.setdefault('attn_implementation', 'eager')
```
akaIDIOT: When all seemed fine, the loaded model would raise a `RuntimeError: could not create a primitive` from `transformers/integrations/sdpa_attention.py`, clanker suggested setting the attention implementation to eager, which actually dodged the error and made the embedder tests succeed 🙏🏻

citatio/pyproject.toml
```diff
@@ -47,6 +47,10 @@ test = [
     "testcontainers[postgres]>=4.14.2",
 ]
 
+[[tool.pdm.source]]
+name = "pytorch-xpu"
+url = "https://download.pytorch.org/whl/xpu"
```
akaIDIOT, line 52: Was hoping this would get magically taken from the `asmtransformers` definitions, but it wouldn't add the xpu stuff when locking. 

Maybe [PDM's new workspace definitions](https://pdm-project.org/en/latest/usage/workspace/) would help here?

akaIDIOT: That's exactly what happens here, that works as intended (though requires PDM 2.28+).

PR #126, opened 2026-08-13
Use pdm workspace

akaIDIOT:


pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[project]
+requires-python = ">=3.13,<3.14"
+
+[dependency-groups]
```
akaIDIOT, line 4: Not a fan of moving these here, but it seemed that they weren't picked up by PDM when left in their respective sub-projects 🤔

